PR #8, opened 2019-01-21
Make `value` and `child` parameters required

weeeBox:


rrousselGit:
Nice catch, looks good! 👌

PR #25, opened 2019-02-20
Idea: Throw error when Provider not found

brianegan:
Right now, the Provider.of method returns `null` if a provider isn't found. 

My guess: This leads to subtle runtime errors, and might leave users wondering what went wrong! This PR changes the `of` method. Now, if no provider is found, an error is immediately thrown (Fail Fast) and gives advice to the developer as to why the provider might not be returning a value.

This same type of code was added to ScopedModel after receiving several error reports when users ran into this exact problem. This error messaging has dramatically cut down on the number of open issues, so I think it's working well!

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -60,10 +60,7 @@ class Provider<T> extends InheritedWidget implements ProviderBase {
             as Provider<T>;
 
     if (provider == null) {
-      throw ProviderNotFoundError(
-        T.toString(),
-        context.widget.runtimeType.toString(),
-      );
+      throw ProviderNotFoundError(T, context.widget.runtimeType);
     }
 
     return provider.value;
@@ -598,10 +595,10 @@ class ValueListenableProvider<T> extends HookProvider<T> {
 /// Widget tree.
 class ProviderNotFoundError extends Error {
   /// The type of the value being retrieved
-  final String valueType;
+  final Type valueType;
 
   /// The type of the Widget requesting the value
-  final String widgetType;
+  final Type widgetType;
 
   /// Create a ProviderNotFound error with the type represented as a String.
   ProviderNotFoundError(
@@ -625,17 +622,6 @@ To fix, please:
 
 If none of these solutions work, please file a bug at:
 https://github.com/rrousselGit/provider/issues
-      ''';
+''';
   }
-
-  @override
-  bool operator ==(Object other) =>
-      identical(this, other) ||
-      other is ProviderNotFoundError &&
-          runtimeType == other.runtimeType &&
-          valueType == other.valueType &&
-          widgetType == other.widgetType;
-
-  @override
-  int get hashCode => valueType.hashCode ^ widgetType.hashCode;
 }
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -58,7 +58,12 @@ class Provider<T> extends InheritedWidget implements ProviderBase {
         ? context.inheritFromWidgetOfExactType(type) as Provider<T>
         : context.ancestorInheritedElementForWidgetOfExactType(type)?.widget
             as Provider<T>;
-    return provider?.value;
+
+    if (provider == null) {
+      throw ProviderNotFoundError();
+    }
+
+    return provider.value;
   }
 
   @override
@@ -584,4 +589,25 @@ class ValueListenableProvider<T> extends HookProvider<T> {
   ///
   /// It is fine to change the instance of [valueListenable].
   final ValueListenable<T> valueListenable;
+}
+
+/// The error that will be thrown if the Provider cannot be found in the
+/// Widget tree.
+class ProviderNotFoundError extends Error {
+  @override
+  String toString() {
+    return '''Error: Could not find the correct Provider.
+    
+To fix, please:
+          
+  * Provide types to Provider<MyClass>
+  * Provide types to Consumer<MyClass> 
+  * Provide types to Provider.of<MyClass>() 
+  * Always use package imports. Ex: `import 'package:my_app/my_code.dart';
+  * Ensure the correct `context` is being used.
+  
+If none of these solutions work, please file a bug at:
+https://github.com/rrousselGit/provider/issues
+      ''';
+  }
 }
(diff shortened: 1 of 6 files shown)


rrousselGit:
Fine by me.
I would enhance the error a bit though by visiting parents and children of the `context` used.

Such that we get the following error:

```dart
Error, no Provider<String> found as ancestor of MyWidget located at:
...
Scaffold
  Center
    MyWidget

____

A Provider<String> have been found in the near descendants of MyWidget. Did you use the wrong BuildContext?

MyWidget
  Center
    Provider<String>
```

brianegan:
Do you think that might get a bit crazy? The example looks great, but it feels like in reality there could be a ton of ancestor Widgets in the tree and the output might be hard to read.

rrousselGit:
There are a few things we can do:

- Filter out widgets that starts by `_`
- Take at most 5 widgets
- Above 5, truncate with a `...`

rrousselGit:
My point is, this is by far the biggest issue newcomers to Flutter face. At least that's what StackOverflow is showing

We need to give as much information as possible

brianegan:
Yep, it's such a common problem. I'll give it a shot to see if it's useful or not!

lib/src/provider.dart
```diff
@@ -58,7 +58,12 @@ class Provider<T> extends InheritedWidget implements ProviderBase {
         ? context.inheritFromWidgetOfExactType(type) as Provider<T>
         : context.ancestorInheritedElementForWidgetOfExactType(type)?.widget
             as Provider<T>;
-    return provider?.value;
+
+    if (provider == null) {
```
rrousselGit: Should this be an assert?

This would fasten a bit the release build. That's what Flutter does with things like `Scaffold.of`

brianegan:
This is the kind of output I get when wrapping a Scaffold inside a Provider and get the 5 most recent :

```
Expected: Provider<String>
Found:
 DefaultTextStyle
   AnimatedBuilder
     CustomMultiChildLayout
       LayoutId
         MediaQuery
```

Not sure what you think, but not sure this is super helpful.

brianegan:
Hrm, not sure what happened to the `assert` comment (showed up in my email), but I'd personally say no -- it should be a regular error.

If this happens in production, you really want to know about it. Silent failures in prod are what lead to terrible Bug reports that are impossible to solve.

rrousselGit:
I think it's important to output the name of the widget associated to `context`. This is more readable than looking into the stacktrace

___


The most useful tree is probably by looking in the next few children, as that's the real source of mistakes and will output something more usable.

Because even with a widget as big as Scaffold like so:

```dart
  return Scaffold(
    body: Provider<String>(),
    floatingButton: Button(onPressed: () => Provider.of(context))
  );
}
```

then we'll get a stack trace as such:

```
MyWidget
  Scaffold
    ...
       Provider<String>
```

rrousselGit:
a depth of 5 may be too small btw.

Also even if in some cases it's unreadable, I think it's still valuable to put this information in – as long as it's at the bottom of the error.


It doesn't hurt anybody and but can be helpful

brianegan:
Interesting -- Unfortunately, I cannot access the current children while building (makes sense!). But: This means I have to delay accessing the children until the build is complete. By that point, any previous children have been replaced by the `ErrorWidget`. So the only output you can receive is:

```
MyWidget
  ErrorWidget
```

Therefore, my current approach looks like this:

```
{
    return '''
Error: Could not find the correct Provider<$_type> above this $_currentWidget Widget 

To fix, please:

  * Ensure the Provider<$_type> is an ancestor to this Widget 
  * Provide types to Provider<$_type>
  * Provide types to Consumer<$_type>
  * Provide types to Provider.of<$_type>()
  * Always use package imports. Ex: `import 'package:my_app/my_code.dart';
  * Ensure the correct `context` is being used.

If none of these solutions work, please file a bug at:
https://github.com/rrousselGit/provider/issues
      ''';
  }
```

rrousselGit:
Oh – sorry about that. Didn't think of it.
I guess we can't do much better then. Thanks for your time!

Tests and we're fine 😄

brianegan:
Just found a couple minutes to patch the remaining tests!

test/provider_test.dart
```diff
@@ -113,6 +109,14 @@ void main() {
       expect(buildCount, equals(2));
     });
 
+    testWidgets('throws an error if no provider found', (tester) async {
+      await tester.pumpWidget(
+        const Builder(builder: Provider.of),
+      );
+
+      expect(tester.takeException(), isInstanceOf<ProviderNotFoundError>());
```
rrousselGit: I think `isInstanceOf` is deprecated

Since we're doing custom stuff with the error message, I think we should test the message too. 

```dart
expect(..., const TypeMatcher<ProviderNotFoundError>()
 .having((err) => err.message, "message", <contains the widget name>);
```

brianegan: Ermagerd, the Test APIs are so hard to use right now as they're transitioning to this new setup :/

Flutter has its own implementation of `TypeMatcher` in the Widgets package, so when I was attempting to use that it wasn't working (because it's not a test matcher).

In addition, `isInstanceOf` isn't deprecated in `flutter_test`, which is why we didn't see any analysis errors on Travis.

That said, it looks like I can do some annoying imports tricks to hide the `TypeMatcher` from the widget framework and use the `TypeMatcher` directly from the matchers package, which gives ya the additional `having `power.

brianegan: Changed the tests. I've implemented a simple `==` function on the Error, which means we don't have to do any of that type matching, we can just say `throwsA(ProviderNotFoundError('valueType', 'widgetType'))`

test/multi_provider_test.dart
```diff
@@ -37,16 +37,7 @@ void main() {
       expect(find.text('Foo'), findsOneWidget);
     });
 
-    testWidgets('MultiProvider with empty providers returns child',
```
rrousselGit: Why did you remove this test?

brianegan: It's a duplicate of the test directly above it.

test/provider_test.dart
rrousselGit: I'm not a fan of using == for this, but it's fine. Thanks :smile:

rrousselGit:
And we're done!

Thanks a lot for your precious time. 😄

brianegan:
Woohoo! Hope it helps the folks using the library :)

PR #27, opened 2019-02-28
v2.0.0

rrousselGit:
- [x] remove the dependency on flutter_hooks.
- [x] add `Flutter Team <flutter-dev@googlegroups.com>` to the authors in `pubspec.yaml`
- [ ] potentially add an optional `child` argument on `Consumer` for performance optimization
- [x] A `ListenableProvider`, like `ChangeNotifierProvider` but for `Listenable`.
- [ ] update documentation/readme/changelog


brianegan:
Woohoo! Cool to see this placeholder & plan :)

  >  potentially add an optional child argument on Consumer for performance optimization

Just my 2 cents: I'm mixed on this one. I feel like I've seen a lot of peeps skip this feature of ScopedModel and it just makes more work for people defining the `builder` function.

If folks need to cache the child, I feel like `const` Widgets are a better option.

rrousselGit:
I agree.

Another thing I usually do is:

```dart
Widget build() {
  final child = SomeWidgetTree();

  return WhateverBuilder(
    builder: (context) {
      return Container(child: child);
   };
}
```

This behaves like the `child` argument. 
But we lose a bit in readability.

brianegan:
Looking really good! One API inconsistency that tripped me up: In some cases, you use a `valueBuilder` to create the value, in other cases you use a `builder` property.

IMO, I think this should be called `valueBuilder` and consistent everywhere. `builder` is so ingrained in my mind as a WidgetBuilder of some kind.

rrousselGit:
Oh nice catch. I missed that one. Will fix 👌

rrousselGit:
Although I'm not too sure about the `valueBuilder` vs `builder`. 
The former is pretty long for not much, and the compiler prevents you from returning widgets.

brianegan:
Maybe I can explain my user journey, and you let me know whatca think:


  - Start creating a ChangeNotifierProvider
  - See the `builder` param
  - Write a function to return a Widget
  - Realize it's meant to build the value
  - Oops, ok I'll use child and make `builder` return the ChangeNotifier.
  
When it comes to providing a `builder` to a Flutter Widget, what comes to mind?

I'm not saying I didn't get IDE help that `builder` was the correct thing to use, but my first instinct was to use that to create a function that returned a Widget. I don't think I'll be alone in making that assumption, given the insane number of Widgets that take a `builder` function to create the Widget.

It's a little longer, but less error-prone.

lib/src/listenable_provider.dart
```diff
@@ -0,0 +1,232 @@
...
+  static void _disposer(BuildContext context, ChangeNotifier notifier) =>
+      notifier?.dispose();
+
+  const ChangeNotifierProvider({
+    Key key,
+    T notifier,
+    Widget child,
+  }) : super(key: key, listenable: notifier, child: child);
+
+  const ChangeNotifierProvider.builder({
```
brianegan: More feedback: I prefer the old `stateful` constructor instead of `builder` for API consistency as well (thinking along the lines of "Ah, this like a StatefulProvider but for ChangeNotifiers!")

rrousselGit: I don't like this one. Because then peoples would request `Provider.stateful`, and ultimately `Provider.whatever` instead of `WhateverProvider` when it doesn't necessarily make sense.

rrousselGit:
That makes sense, and it also reflects the typedef name (which is named`ValueBuilder<T>`).

But at the same time, the mistake happens only one and it makes all subsequent usage slightly easier.
Also, peoples copy-pasting code from the readme to try it out won't have the issue. 

I supposed we can let others tell us what they think (@jogboms @filip ?)

filiph:
I don't mind either way. I wonder if requiring the `child` param would make the problem go away. AFAIK, there's no real reason to have a StatefulProvider without a child (maybe a failure of my imagination). If `child` is required, my IDE adds that additional nudge -- this is wrong and you should add child.

Another option is `buildValue` instead of `builder` or `valueBuilder`.

Sorry, not very helpful, I know.

lib/src/listenable_provider.dart
filiph: I, for one, slightly prefer "builder" over "stateful". My main reason is accessibility for newcomers. Stateful seems like implementation-leaking. What the user cares about is that they'll build a value to be provided. See a top-level comment below for a more general discussion about builder/stateful.

filiph:
Do we have any usage stats in regards to `*Provider.stateful` from 1.0? I would like to check the hypothesis that most users, most of the time, use `*Provider.stateful` (or `*Provider.builder`) instead of the default constructor. If that's the case, and considering the fact that non-stateful providers can lead to confusion, I wonder if it makes sense to switch it around, so that what is now `builder` becomes the default constructor, and vice versa. Keeping in mind Remi's design goal of not letting people shoot themselves in the foot too often.

I will do some light github searching later today.

filiph:
Update: github search is evil. Not much luck with google search, either. If my search brought anything, it's that `StatefulProvider` is not significantly more used that `Provider`. So my [previous comment](https://github.com/rrousselGit/provider/pull/27#issuecomment-475731858) is naught. Unless there is a strong opinion that StatefulProvider is what people really want but don't realize it, I think there's nothing to do.

rrousselGit:
There is actually no `Provider.stateful` in 1.x. The only one being `StatefulProvider`, but it's not a named constructor

Users are definitely fallbacking to `Provider` instead of `StatefulProvider` – which is not a good thing.

I've been thinking of merging `StatefulProvider` and `Provider` to make _builder_ the default behavior.
But this requires to do some dirty tricks to work as `Provider` is an InheritedWidget but `StatefulProvider` a `StatefulWidget` – and I don't like the idea of promoting bad practices.

rrousselGit:
IMO this not `provider` specific anyway. We should have a custom linter rule for flutter that warns if we're instantiating anything that is not immutable inside _build_ method

Other widgets suffer from the same issue:

```dart
FutureBuilder(
  future: doSomeHttpRequest(),
  builder: ...
)
```

filiph:
Gotcha.

> But this requires to do some dirty tricks to work as Provider is an InheritedWidget but StatefulProvider a StatefulWidget – and I don't like the idea of promoting bad practices.

Since this might be the last chance to do that change, I would at least consider having:

* `Provider` — the widget previously known as "StatefulProvider", the default
* `OneShotProvider` (sorely needs a better name) — the widget previously known as "Provider"

Of course, this is a big breaking change. And I get your comment that this is far from being a problem of just `package:provider`. I'm just trying to prevent onboarding with things like "There's Provider, but you almost never want to use it, instead use StatefulProvider and you'll be safer." That automatically begs the question why the former has a simpler name than the latter.

No strong opinion here. Just throwing this out.

rrousselGit:
>  I wonder if requiring the child param would make the problem go away. AFAIK, there's no real reason to have a StatefulProvider without a child

That's not possible because of `MultiProvider`:


```
MultiProvider(
  providers: [
    StatefulProvider(valueBuilder: (context) => Class()),
  ],
  child: Foo(),
)
```

filiph:
> That's not possible because of MultiProvider:

Oh! 🤦‍♂️

rrousselGit:
I think it's fine as is.

`StatefulProvider` is not perfect either and I don't think we should recommend to use it all the time.
Peoples are tempted to use parameters inside the builder of `StatefulProvider`:

```dart
int foo;

StatefulProvider(
  builder: (context) => foo * 2,
)
```

This is likely to be undesired because updating `foo` won't update the value provided. 

That's the reason why I added the dependency on `flutter_hooks`. `HookProvider` do what `StatefulProvider` can but without that update issue.

rrousselGit:
A few other things I've been considering:

- renaming _StatefulProvider_ to _MemoizedProvider_

_Stateful_ in Flutter is used to create a mutable state. _StatefulProvider_ does the opposite, it creates an immutable state, and _Provider_ is the one that allows mutation.
_Memoized_ may have more explicit intent. 

- creating an _AnimatedProvider_ – a _Provider_, with a _Duration_, _Curve_ and a `Tween<T> Function(T begin, T end, double t)` 

While this is purely additive, it reinforces the idea of **not** renaming _Provider_.

filiph:
Makes sense, thanks!

brianegan:
`MemoizedProvider` is great.

In that case, would it make sense to have `ChangeNotifierProvider.memoized` instead of `ChangeNotifierProvider.builder`? I know you weren't a fan of the suggestion the first time, but I think `memoized` describes the functionality a bit better than `builder` without leaking implementation details, and would provide a more consistent API if you've seen / used `MemoizedProvider`.

rrousselGit:
> In that case, would it make sense to have ChangeNotifierProvider.memoized

I agree.
But then it really bothers me that we have a `MemoizedProvider` instead of `Provider.memoized`

brianegan:
Could `Provider.memoized` be a `static` factory that returns a `_MemoizedProvider`?

rrousselGit:
That won't do. It means that `Provider` and `Provider.memoized` have a different `runtimeType`.

Which also means that Flutter would destroy the subtree when changing between the default and `memoized` constructor – whatever the reason is.

The only way is to make the current `Provider` private and always wrap it into a `StatefulWidget` that can switch between the default and _memoized_ constructor (like how other providers works).

But I'm not a fan of adding overhead to the primitive `Provider`. Especially when I have an architecture – that I want to make an article on when 2.0 is out – that extensively use `Provider` directly.

elforce001:
Hi @rrousselGit . I came here because of @brianegan recommended this library. Do we start using this version (V2.0.0) or just go to the master branch instead?

enetor:
+1 to renaming StatefulProvider as MemoizedProvider.

filiph:
FWIW, I also like `MemoizedProvider` about as much as `StatefulProvider`. 

`StatefulProvider` is confusing, but at least it's not introducing a completely new term that is heavily associated with a slightly different concept (that of memoizing pure functions). Also — considering `Stateful/MemoizedProvider` is going to be used a lot — it's a mouthful.

I don't want to keep this from landing, but allow me a quick brainstorm. Feel free to ignore this.

* `Provider` - `StatefulProvider` (current)
* `Provider` - `MemoizedProvider`
* `Provider` - `BuilderProvider`
* `ValueProvider` - `Provider`
* `StatelessProvider` - `Provider`

rrousselGit:
if `MemoizedProvider` is too long, would `MemoProvider` work?

That's the naming used by React. Although I don't like it

brianegan:
Just my 2 cents, but I don't think `MemoizedProvider` is too long. Tons of classes / constructors in Flutter are about that same character length, just off the top of my head:

  - `TabBarController`
  -  `ChangeNotifier`
  - `ListView.builder`
  - `AnimationController`
  - `SlideTransition`

filiph:
FWIW, I don't think length is an issue, neither. Here are the 2 problems I see with the name:

* memoization is strongly associated with performance optimization and caching, of function results. (Wikipedia: "memoisation is an optimization technique used primarily to speed up computer programs by storing the results of expensive function calls and returning the cached result".) In my opinion, it's a little bit of a stretch to apply it to the concept of "stateful" in Flutter.
  * Even in React, `memo` is a function, and the [documentation](https://reactjs.org/docs/react-api.html#reactmemo) makes it very clear this is about performance optimization ("This method only exists as a performance optimization. Do not rely on it to “prevent” a render, as this can lead to bugs.")
* memoization is a technical term not well known outside CS. Flutter naming is often sticking to simple words (`Column` and `Row` instead of something like Vertical/Horizontal, `builder` instead of factory, `Notifier` instead of pubsub). This is not a holy cow, but it's nice to be able to use simple, non-jargony words to talk about these complex topics. I personally have no problem with memoization, and obviously nobody else here neither, but I worry that a regular developer might. Even just reading about `React.memo`, I see things like "If you don’t have a hardcore computer science background, like me, it may throw you off a little bit." and "not even Wikipedia knows how to spell memoization". These, to me, are indications that memoization is not a word that people readily understand.

Just my 2 cents.

rrousselGit:
> memoization is a technical term not well known outside CS

Agreed on that one.


What about `CachedProvider` then?

___


Another alternative is to add a way to update the state of "StatefulProvider"

I think the hook API is a good fit (without the dependency of course).

We could have:

```dart
String foo;

StatefulProvider(
  builder: (_) => Class(foo), 
  dispose: (_, o) => o.dispose(),
  keys: [foo], 
  child: Foo(),
)
```

Which means that whenever something inside `keys` change, the previous value is disposed and a new one is created.

This makes it more _stateful_ as we have a mean to update the state.

Hixie:
@filiph asked me to take a look at this.

I think `Provider` for the version with the "create" and "dispose" callbacks makes a lot of sense. I think `ValueProvider` for the version with the value makes sense; I could also imagine having it just be the same constructor but you require that either the callbacks be non-null _or_ the value be non-null but not both. Or a named constructor `Provider.fromValue` or something.


On another note:

> Another thing I usually do is:
> 
> ```dart
> Widget build() {
>   final child = SomeWidgetTree();
> 
>   return WhateverBuilder(
>     builder: (context) {
>       return Container(child: child);
>    };
> }
> ```
> 
> This behaves like the `child` argument.

(I'm assuming the `Container` here is just a placeholder for something more complicated. Obviously you would never need to just say `Container(child: child)`.)

I agree that the above is basically equivalent to the `child` argument. We've found historically that having the explicit `child` argument is good because it gives somewhere for us to hammer in the idea that you want to do this for good performance rather than rebuilding the child each time.

brianegan:
Nice one! The API is looking really clean and consistent now.

rrousselGit:
Cool! I was about to ask the same thing

> I think this is ready to be published to pub as 2.0.0-alpha 

It is already flagged as `2.0.0-dev`. Should it be changed to `alpha`?

filiph:
Ah, dev is totally fine.

>

jiaeric:
Hey, sorry for jumping into this so late.
Do you think ListenableProvider is needed? It does add some complexity, where it might be cleaner on the consumer side

A ListenableConsumer can just use an ordinary provider, and look like this:

`class ListenableConsumer<T extends Listenable> extends StatelessWidget {
  final Widget Function(BuildContext context, T value, Widget child) builder;

  ListenableConsumer({Key key, @required this.builder}) : super(key: key);

  @override
    return Consumer<T>(
        builder: (context, value) =>
            AnimatedBuilder(animation: value, builder: builder));
  }
}`

rrousselGit:
@jiaeric  I moved that question to #39 (as this MR is already merged).

pubspec.yaml
```diff
@@ -10,7 +10,6 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  flutter_hooks: ">=0.2.1 <1.0.0"
```
jogboms: Oh! I just found it when you removed it.

PR #28, opened 2019-03-05
Remove flutter hooks

rrousselGit:


example/stateful_provider.dart
```diff
@@ -25,7 +25,7 @@ class MyApp extends StatelessWidget {
   Widget build(BuildContext context) {
     return StatefulProvider<Bloc>(
       valueBuilder: (_) => Bloc(),
-      onDispose: (_, value) => value.dipose(),
+      dispose: (_, value) => value.dipose(),
```
jogboms: We would need to update the README after this as well.

rrousselGit: Right. There's a lot of documentation update needed

I'll fix tests first though, are most needs a rewrite.

rrousselGit: I'll fix that in another PR related to documentation

PR #34, opened 2019-03-26
Improve the overal styling and documentation of provider.dart.

enetor:


lib/src/provider.dart
```diff
@@ -202,32 +215,16 @@ class StatefulProvider<T> extends StatefulWidget
   /// Allows to specify parameters to [StatefulProvider]
   StatefulProvider({
     Key key,
-    @required this.valueBuilder,
     this.child,
+    @required this.valueBuilder,
```
filiph: [nit] I think it's good style to start with `@required` paramaters.

enetor: done!

lib/src/provider.dart
```diff
@@ -201,33 +214,17 @@ class StatefulProvider<T> extends StatefulWidget
...
-
-  /// [dispose] is a callback called when [StatefulProvider] is
-  /// removed for the widget tree, and pass the current value as parameter.
-  /// It is useful when the provided object needs to have a custom dipose behavior,
-  /// such as closing streams.
   final Disposer<T> dispose;
-
-  /// The widget below this widget in the tree.
-  ///
-  /// {@macro flutter.widgets.child}
```
rrousselGit: Out of curiosity, why remove these pieces of documentation?

enetor: Mostly because the current documentation is redundant - the class-level comment has good coverage. I can keep them if you prefer though. I don't have a strong opinion here.

enetor: I actually have a question here:

These members are public - is this necessary? I see that some of the test code accesses them. In my opinion these fields can as well be marked package-private (with the underscore prefix). WDYT?

rrousselGit: Documenting every single public field is necessary, otherwise pub decreases the health score of the package.

> these fields can as well be marked package-private

I'm not sure. 
I've never seen a widget hiding its fields.

enetor: Added comments for those fields.

I've seen them :-)

Nevertheless, this is not a big issue. And I do not want to block this change.

lib/src/provider.dart
```diff
@@ -238,7 +238,7 @@ class StatefulProvider<T> extends StatefulWidget
   /// parameter.
   ///
   /// This function is useful when the provided value has custome disposal
```
rrousselGit: Did I seriously write "custome"? 🤦‍♂️

lib/src/provider.dart
```diff
@@ -238,7 +238,7 @@ class StatefulProvider<T> extends StatefulWidget
...
-  /// behavior. For example a [Stream] which needs to be closed explicitly.
+  /// behavior. For example a stream which needs to be closed explicitly.
```
rrousselGit: I think we should change `stream` to `sink`

enetor: done

lib/src/provider.dart
enetor: nope that was me...

rrousselGit:
Is this ready to be merged? 😄

enetor:
I'll squash the commits and update here.

enetor:
Hi Remi, this pull request is now ready to be merged. Thanks!

enetor:
(Clicked the wrong button...)

PR #35, opened 2019-03-28
Merge Provider and StatefulProvider

rrousselGit:
This fuses `Provider` and `StatefulProvider` using named constructors.

Such that we have:

```dart
Provider<T>(
  builder: (_) => T(),
  dispose: (_, t) => t.dispose(),
  child: Foo(),
)
```

and

```dart
Provider<T>.value(
  value: t,
  child: Foo(),
)
```

____

This also factorized the code to remove duplicates (as this swap between value/builder constructor is very common) and normalized providers to use the same logic:

- the default constructor is a builder
- providers offer a `value` named constructor


rrousselGit:
I'm still not quite fond of always wrapping the _InheritedWidget_ into a _StatefulWidget_. But API wise, It's probably the best approach

jogboms:
BREAKING CHANGE!!

Please increase the major version to reflect this before publishing.

rrousselGit:
> Please increase the major version to reflect this before publishing.

This is for the v2 😄

lib/src/adaptative_builder_widget.dart
```diff
@@ -0,0 +1,92 @@
+import 'package:flutter/widgets.dart';
+
+/// A function that creates an object of type [T].
+typedef ValueBuilder<T> = T Function(BuildContext context);
+
+abstract class AdaptativeBuilderWidget<T, B> extends StatefulWidget {
```
enetor: Do you mean "Adaptive"? Not sure if "Adaptative" is a valid word

lib/src/adaptative_builder_widget.dart
```diff
@@ -0,0 +1,92 @@
...
+  const AdaptativeBuilderWidget({
+    Key key,
+    @required this.builder,
```
enetor: filip mentioned it's preferred to put 'required' arguments before others

lib/src/adaptative_builder_widget.dart
rrousselGit: Fixed
I confused the French and English word. 🙃

lib/src/adaptative_builder_widget.dart
rrousselGit: Is this really what's wanted for widgets?

I saw somewhere else a statement where keys should always be first and children last – and a quick look through Flutter seems to confirm this (with some exceptions).

filiph: No big deal in any case. There's a lint for this, which would fire for the code above, but that's obviously optional. Consider it a nit and go with whatever works best for this codebase.

example/stateful_provider.dart
```diff
@@ -23,8 +23,8 @@ class Bloc {
 class MyApp extends StatelessWidget {
   @override
   Widget build(BuildContext context) {
-    return StatefulProvider<Bloc>(
-      valueBuilder: (_) => Bloc(),
+    return Provider<Bloc>(
```
filiph: [nit] Can we change the names of these files? Since the "default" is now `Provider()` (and not `Provider.value()`) I'd be for either merging the two files or doing something like `provider.dart` (for stateful)` and `provider_value.dart` (for `Provider.value()`).

lib/src/provider.dart
```diff
@@ -202,38 +152,61 @@ UpdateShouldNotify<T> debugGetProviderUpdateShouldNotify<T>(
...
 /// ```
-class StatefulProvider<T> extends StatefulWidget
+class Provider<T> extends AdaptiveBuilderWidget<T, T>
     implements SingleChildCloneableWidget {
-  /// Allows to specify parameters to [StatefulProvider]
-  StatefulProvider({
-    @required this.valueBuilder,
+  /// Obtains the nearest Provider<T> up its widget tree and returns its value.
+  ///
+  /// If [listen] is true (default), later value changes will trigger a new
```
filiph:
```
If [listen] is `true` ...
```

lib/src/provider.dart
```diff
@@ -202,38 +152,61 @@ UpdateShouldNotify<T> debugGetProviderUpdateShouldNotify<T>(
...
 ///   }
 /// }
 /// ```
-class StatefulProvider<T> extends StatefulWidget
+class Provider<T> extends AdaptiveBuilderWidget<T, T>
     implements SingleChildCloneableWidget {
-  /// Allows to specify parameters to [StatefulProvider]
-  StatefulProvider({
-    @required this.valueBuilder,
+  /// Obtains the nearest Provider<T> up its widget tree and returns its value.
```
filiph: [nit] Put `Provider<T>` into backticks.

test/listenable_provider_test.dart
```diff
@@ -9,36 +9,66 @@ import 'common.dart';
...
         providers: [
-          ListenableProvider(listenable: listenable),
+          ListenableProvider.value(listenable: listenable),
         ],
         child: Container(key: key),
       ));
 
       expect(Provider.of<ChangeNotifier>(key.currentContext), listenable);
     });
+    test('works with MultiProvider #2', () {
```
filiph: [nit] a better test name would be great

test/multi_provider_test.dart
```diff
@@ -63,8 +63,12 @@ void main() {
 
       expect(find.text('Foo'), findsOneWidget);
 
-      // p1 cannot access to /p2/p3
-      expect(Provider.of<int>(k1.currentContext), 42);
```
filiph: (Interesting. I'm surprised this worked.)

filiph:
👍

PR #38, opened 2019-04-12
documentation

rrousselGit:
Filling the gap in documentation

### DIFF
--- a/lib/src/listenable_provider.dart
+++ b/lib/src/listenable_provider.dart
@@ -3,15 +3,20 @@ import 'package:flutter/material.dart';
 import 'package:provider/src/adaptive_builder_widget.dart';
 import 'package:provider/src/provider.dart';
 
+/// Listens a [Listenable], expose it to its descendants
+/// and rebuilds dependents whenever the listener emits an event.
+///
+/// See also:
+///   * [ChangeNotifierProvider], a subclass of [ListenableProvider] specific to [ChangeNotifier].
+///   * [Listenable]
 class ListenableProvider<T extends Listenable>
     extends AdaptiveBuilderWidget<T, T> implements SingleChildCloneableWidget {
-  const ListenableProvider.value({
-    Key key,
-    @required T listenable,
-    this.child,
-  })  : dispose = null,
-        super.value(key: key, value: listenable);
-
+  /// Creates a [Listenable] using [builder] and subscribes to it.
+  ///
+  /// [dispose] can optionally passed to free resources
+  /// when [ListenableProvider] is removed from the tree.
+  ///
+  /// [builder] must not be `null`.
   const ListenableProvider({
     Key key,
     @required ValueBuilder<T> builder,
@@ -19,7 +24,27 @@ class ListenableProvider<T extends Listenable>
     this.child,
   }) : super(key: key, builder: builder);
 
+  /// Listens [listenable] and expose it to all of [ListenableProvider] descendants.
+  ///
+  /// Rebuilding [ListenableProvider] without
+  /// changing the instance of [listenable] will not rebuild dependants.
+  const ListenableProvider.value({
+    Key key,
+    @required T listenable,
+    this.child,
+  })  : dispose = null,
+        super.value(key: key, value: listenable);
+
+  /// Function used to dispose of an object created by [builder].
+  ///
+  /// [dispose] will be called whenever [ListenableProvider] is removed from the tree
+  /// or
   final Disposer<T> dispose;
+
+  /// The widget that is below the current [ListenableProvider] widget in the
+  /// tree.
+  ///
+  /// {@macro flutter.widgets.child}
   final Widget child;
 
   @override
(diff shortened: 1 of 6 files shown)


lib/src/listenable_provider.dart
```diff
@@ -3,23 +3,48 @@ import 'package:flutter/material.dart';
 import 'package:provider/src/adaptive_builder_widget.dart';
 import 'package:provider/src/provider.dart';
 
+/// Listens a [Listenable], expose it to its descendants
```
brianegan: "Listens to a [Listenable"

lib/src/listenable_provider.dart
```diff
@@ -3,23 +3,48 @@ import 'package:flutter/material.dart';
...
+/// and rebuilds dependents whenever the listener emits an event.
+///
+/// See also:
+///   * [ChangeNotifierProvider], a subclass of [ListenableProvider] specific to [ChangeNotifier].
```
brianegan: Add `ValueListenableProvider` to this list?

lib/src/listenable_provider.dart
```diff
@@ -3,23 +3,48 @@ import 'package:flutter/material.dart';
...
+    Key key,
+    @required T listenable,
+    this.child,
+  })  : dispose = null,
+        super.value(key: key, value: listenable);
+
+  /// Function used to dispose of an object created by [builder].
+  ///
+  /// [dispose] will be called whenever [ListenableProvider] is removed from the tree
+  /// or
```
brianegan: Leftover or incomplete sentence?

lib/src/listenable_provider.dart
```diff
@@ -107,27 +132,33 @@ class _ListenableProviderState<T extends Listenable>
   }
 }
 
+/// Listens a [ChangeNotifier], expose it to its descendants
```
brianegan: Listens to a [ChangeNotifier]

lib/src/listenable_provider.dart
```diff
@@ -107,27 +132,33 @@ class _ListenableProviderState<T extends Listenable>
...
+/// See also:
+///   * [ListenableProvider], similar to [ChangeNotifier] but works with any [Listenable].
+///   * [ChangeNotifier]
 class ChangeNotifierProvider<T extends ChangeNotifier>
     extends ListenableProvider<T> implements SingleChildCloneableWidget {
   static void _disposer(BuildContext context, ChangeNotifier notifier) =>
       notifier?.dispose();
 
-  const ChangeNotifierProvider.value({
+  /// Create a [ChangeNotifier] using [builder] function and automatically dispose it
```
brianegan: using the [builder]

lib/src/provider.dart
```diff
@@ -144,6 +144,10 @@ class MultiProvider extends StatelessWidget
 /// The following example instantiates a `Model` once, and disposes it when
 /// [Provider] is removed from the tree.
 ///
+/// {@template provider.updateshouldnotify}
+/// [updateShouldNotify] can optionally be passed to avoid unnecessaryly rebuilding dependants when nothing changed.
+/// Defaults to `(previous, next) => previous == next`. See [InheritedWidget.updateShouldNotify] for more informations.
```
brianegan: is the default `previous == next` or `previous != next`? Seems weird to notify if the two are equal, but not when the are different?

lib/src/stream_provider.dart
```diff
@@ -6,9 +6,30 @@ import 'package:provider/src/provider.dart';
 
 typedef ErrorBuilder<T> = T Function(BuildContext context, Object error);
 
+/// Listens to a [Stream<T>] and exposes [T] to its descendants.
+///
+/// It is considered an error to pass a stream that can emit errors without providing
+/// a [orElse] method.
```
brianegan: an [orElse]

lib/src/stream_provider.dart
```diff
@@ -6,9 +6,30 @@ import 'package:provider/src/provider.dart';
...
+///
+/// {@macro provider.updateshouldnotify}
+///
+/// See also:
+///   * [Stream]
+///   * [StreamController], to create a [Stream]
 class StreamProvider<T>
     extends AdaptiveBuilderWidget<Stream<T>, StreamController<T>>
     implements SingleChildCloneableWidget {
+  /// Creates a [StreamController] from [builder] and subscribes to it.
```
brianegan: Just wondering why this doesn't require the user to return a `Stream<T>`? Feels weird to have a `StreamProvider` require you return a `StreamController`. I figured the builder would be a function that returns a `Stream`?

lib/src/stream_provider.dart
```diff
@@ -27,9 +49,24 @@ class StreamProvider<T>
...
 
+  /// {@macro provider.streamprovider.initialdata}
   final T initialData;
+
+  /// The widget that is below the current [StreamProvider] widget in the
+  /// tree.
+  /// {@macro flutter.widgets.child}
   final Widget child;
+
+  /// Optional function used whenever the listened [Stream] emits an error.
```
brianegan: "whenever the [Stream] emits an error"

lib/src/stream_provider.dart
```diff
@@ -27,9 +49,24 @@ class StreamProvider<T>
...
+  ///
+  /// [orElse] will be called with the emitted error and
+  /// is expected to return a fallback value without throwing.
+  ///
+  /// The returned value will then be exposed to the descendants of [StreamProvider]
+  /// like any valid value.
```
brianegan: Just wondering if this should be called `onError` or `catchError` or something with `error` in the name? `orElse` sounds a bit weird here.

lib/src/stream_provider.dart
rrousselGit: It's similar to how `builder` from `ValueListenableProvider` returns a `ValueNotifier` instead of `ValueListenable`.

That's to free the resources with `.close()`/`.dispose()`.

If we have a 
```dart
StreamProvider(
  builder: (_) => StreamController().stream,
  ...
)
```
Then even with a `dispose` callback, there's no way to close the controller since its reference is lost.

lib/src/stream_provider.dart
rrousselGit: I like `catchError`. 👌

lib/src/stream_provider.dart
brianegan: Not sure I'm convinced by this... Why would I need to create a StreamController inside the `builder`?

This is kinda how I thought the Widget would work when I first saw the name:

```
StreamProvider<QuerySnapshot>(
  builder: (_) => Firestore.instance.collection('books').snapshots(),
  ...
)
```

or

```
StreamProvider<List<Todo>>(
  builder: (_) => myBloc.todos,
  ...
)
```

or

```
StreamBuilder<String>(
  builder: (_) async* {
    yield "Loading"
    yield await fetchSomething();
  },
  ...
)
```

These would all be pretty awkward with the current API. You'd have to do something like this with all of them:

```
StreamProvider<QuerySnapshot>(
  builder: (_) => StreamController()..addStream(Firestore.instance.collection('books').snapshots()),
  ...
)
```

Overall, I get the need to release resources, but when it comes to Streams, that means cancelling a subscription, not disposing of a controller?

rrousselGit: In the examples you gave you'd use the value constructor:

```dart
StreamProvider.value(
  stream: Bloc.of(context).foo,
 ...
) 
```

Which takes a stream directly.


Canceling subscription/removing listeners is needed for both constructors. But only the builder constructor requires to call `close`/`dispose`

brianegan: Hrm, I thought the `builder` would be most useful to help peeps from recreating Streams over and over in the build method (the Firestore example would suffer from this using the `value` factory), rather than as a way to create your own `StreamController` that would be disposed.

rrousselGit: Hum, I get that.
At the same time I wanted to provide the most harmless solution possible, even if it limits the possibilities. 
I don't want folks to start misusing it and have memory leaks all over the place.


I mean with your suggestion we can technically do:

```dart
StreamBuilder(
  builder: (_) async* {
    var controller = StreamController();
    try {
      yield* controller.stream;
    } finally {
      controller.close();
    } 
  }
)
```

But that's pretty advanced

rrousselGit: For now I'll merge it. I'll open an issue to see if we can improve that aspect

PR #40, opened 2019-04-30
remove updateShouldNotify parameter from Provider.

rrousselGit:
This removed the parameter `updateShouldNotify` from `Provider` constructor (it is still available using `Provider.value`).

It made no sense to be able to pass an `updateShouldNotify` since, by principle, the builder constructor of `Provider` will always provide the same value.


filiph:
LGTM

PR #47, opened 2019-05-01
Add child parameter to Consumer

rrousselGit:


lib/src/consumer.dart
```diff
@@ -5,6 +5,10 @@ import 'package:provider/src/provider.dart';
 /// Obtain [Provider<T>] from its ancestors and pass its value to [builder].
 ///
 /// [builder] must not be null and may be called multiple times (such as when provided value change).
+///
+/// ## Performances optimizations:
```
filiph: typo: s/Performances/Performance/

lib/src/consumer.dart
```diff
@@ -1,64 +1,104 @@
...
   ///
   /// Must not be null.
-  final Widget Function(BuildContext context, T value) builder;
-
-  /// Consumes a [Provider<T>]
-  Consumer({Key key, @required this.builder})
-      : assert(builder != null),
-        super(key: key);
+  /// {@endtemplate}
+  final Widget Function(BuildContext context, T value, Widget child) builder;
```
filiph: Any reason for this particular sorting? ScopedModel has [context, child, value](https://pub.dartlang.org/documentation/scoped_model/latest/scoped_model/ScopedModelDescendantBuilder.html), so having the same ordering would make some people's life easier. Not just when upgrading from one to another, but also when developing — I personally have muscle memory for context, child, value. 

It also corresponds better with Flutter's [TransitionBuilder](https://docs.flutter.io/flutter/widgets/TransitionBuilder.html).

On the other hand, [ValueWidgetBuilder](https://docs.flutter.io/flutter/widgets/ValueWidgetBuilder.html) has the same ordering as you, and it's the closest to `Consumer` that we already have in Flutter. It's just less known than ScopedModel.

I don't have any preference. "context, child, value" for historic purposes, "context, value, child" for more consistency with the framework itself. Make your pick. Just making sure we're intentional about this.

rrousselGit: It follows the "optional parameters are placed after required parameters" rule.
While the typing system of Dart doesn't support it yet, technically `child` is optional.

PR #58, opened 2019-05-11
fix documentation

krusek:
Still a few leftover places where the wrong constructor is called for `Provider`.


rrousselGit:
Actually, I need to push a bug-fix, so I'll just push this PR with it too.

PR #59, opened 2019-05-12
Proxy provider

### LINKED ISSUE #46: [RFC] Simplifying providers that depends on each others
From experience looking at larger Flutter applications, it is relatively common to have some dependency between providers.

For example, we may have a `Provider<Configuration>` and a `Provider<AuthentificationService>` where the latter depends on the former.

But that's currently not something we can represent purely using `provider` package. Our only solution right now is to somehow _eject_ `provider` and handle everything ourselves like so:

```dart
Configuration configuration;

  return MultiProvider(
    providers: [
      Provider.value(value: configuration),
      Provider(builder: (_) => AuthentificationService(configuration)),
    ],
    child: MaterialApp(...),
  );
}
```

That's not good.
It reintroduces the ability to make a circular object graph & forget to update/dispose of an object.
We're loosing all the fancy securities that using widgets for DI brings.

____

The discussion on https://github.com/rrousselGit/provider/issues/44 made me realize that an advanced `Consumer` may be able to solve this issue. 

Naively we could think of doing:

```dart
MultiProvider(
  providers: [
    Provider<Foo>(builder: (_) => Foo()),
    Consumer<Foo>(builder: (context, foo, child) {
      final bar = Bar(foo);
      return Provider<Bar>.value(value: bar, child: child);
    }),
  ].
  child: MaterialApp(...),
);
```

... The problem is that we're doing `final bar = Bar(foo);` inside the build method of `Consumer`.
That's not good either – we may have memory leaks.
But that's something we can fix using a more advanced `Consumer`, which I'll call `ProxyProvider` for now.

The idea behind `ProxyProvider` is that it would work similarly to the default constructor of `Provider`, but also like `Consumer` as it would read values from other providers and pass them to the `builder`.

The same example using `ProxyProvider` would be:

```dart
MultiProvider(
  providers: [
    Provider<Foo>(builder: (_) => Foo()),
    ProxyProvider<Foo, Bar>(builder: (context, Foo foo, Bar previousBar) {
      return Bar(foo);
    }),
  ].
  child: MaterialApp(...),
);
```

In that example, the `builder` of `ProxyProvider` would be called on the first build; or whenever `Foo` changes.

rrousselGit:
closes #46


yaymalaga:
Are you planning to add for the initial release of ProxyProvider the syntax you mentioned in FilledStack pull request discussion?

`ProxyProvider<Dependency, Output>(
  initialBuilder: (_) => Output(),
  builder: (_, dependency, output) => output..dependency = dependency;
  dispose: (_, output) => output.dispose(),
)`

rrousselGit:
It's already added :smile:

I need to add the missing tests, then I'll make a preview release

PR #60, opened 2019-05-12
fix build/listening issue

rrousselGit:


rrousselGit:
cc @filiph 
I'll push a hot-fix.
In some situations, it was possible for `ListenableProvider`/`ChangeNotifier` to **not** notify its dependent that something changed.

This issue happened if the provider rebuilt (with a `setState`) in the same frame where a `Listenable` called its listener for the first time.

PR #80, opened 2019-05-21
Fix broken doc formatting

timsneath:
The code sample here isn't being formatted correctly (see https://pub.dev/documentation/provider/latest/provider/Provider-class.html). I think this change should fix it.


rrousselGit:
Nice catch indeed!

PR #95, opened 2019-05-29
v3.0.0

### LINKED ISSUE #77: [RFC] Should we make a v3.0.0 with name tweaks/constructors swap?
There are a few inconsistencies among the public API

- `StreamProvider.builder` from the default constructor requests a `StreamController` instead of a stream. It'd be more logical for the default constructor to request a `Stream`, and have a named constructor or another class for the `StreamController`.

- `XXProvider.value` have inconsistencies in the name used for the value. `ChangeNotifierProvider` requests a `notifier`, `Provider` requests a `value`, `ValueListenableProvider` requests a `valueListenable`. But they probably should all be named `value`.

- Not an immediate issue, but `ProxyProvider.custom` has a weird naming issue. It takes both a `builder` and a `providerBuilder` argument. Maybe we should globally rename `builder` to something better (`valueBuilder` like it originally was?)

These are minor changes but requires a breaking change.

cc @brianegan @filiph

rrousselGit:
- [x] Allow creators to easily make their own providers (https://github.com/rrousselGit/provider/pull/92 –  providers can no more be instantiated with `const`)
- [x] Unify the naming of `XXProvider.value`: `ChangeNotifierProvider.value(notifier: foo)` -> `ChangeNotifierProvider.value(value: foo)`
- [x] `StreamProvider`'s default constructor should build a stream instead of a stream-controller.
- [x] `StreamProvider` should expose a named constructor that builds a stream-controller automatically disposed.
- [x] initial release of `ProxyProvider`
- [x] initial release of `FutureProvider`.
- [x] Add an health check to unsure that `Provider` is not used instead of `ListenableProvider`

closes #77


rubensdemelo:
It is possible to ship this new version with examples? I would like to help (if needed).

rrousselGit:
What kind of example do you have in mind?

In the dartdoc like so?

```dart
/// Hello
///
/// ```dart
/// Provider(builder: (context) => Foo()),
/// ```
class Provider  {
}
```

Or an actual app?

The latter should probably be in a standalone repository. For the former, feel free to make a PR :smile:

FilledStacks:
> It is possible to ship this new version with examples? I would like to help (if needed).

@rrousselGit @rubensdemelo When proxy provider is released it's a first priority for me to make a tutorial on it. I'll build the same app as I did with the other architecture using provider only without get_it. I'll create a standalone repo outside of my tutorials repo that can be pointed to as one of the examples. 

If you approve of the implementation of course 😁

rrousselGit:
@FilledStacks cool!

If that example requires my aprovement, ideally it should be merged on this repository. 
I refactored the folder architecture, such that we can add a `sample/` folder for examples.

If it's in your priorities, you can start working on it already by forking this branch and using a relative path to make the example depend on `provider`.

FilledStacks:
@rrousselGit you're right! I'll schedule it for next week's tutorial and see if I can get a good architecture going for a basic app. I'll do a pull request if I'm happy with the code. If not I'll just update here.

PR #105, opened 2019-06-01
Prevent Provider from using Listenable/Stream.

### LINKED ISSUE #102: ChangeNotifierProvider vs Provider lint?
Someone just showed me `provider` code that didn't work for them — things were getting updated internally, the `ChangeNotifier` was calling `notifyListeners`, the widget was correctly wrapped in a relevant `Consumer`, there were no static or dynamic errors, but nothing changed in the UI.

The culprit, which took me several minutes to realize, was using `Provider` instead of `ChangeNotifierProvider`.

I don't think there is an easy solution for this, but I thought maybe others would have ideas.

rrousselGit:
This is a health check to ensure that the listenable objects are indeed listened.

closes #102

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/widgets.dart';
 import 'package:provider/src/delegate_widget.dart';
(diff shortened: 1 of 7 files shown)


rrousselGit:
cc @filiph 
Do you think that this is a reasonable solution for #102?

filiph:
That's a great solution, thanks!

lib/src/provider.dart
```diff
@@ -223,6 +205,74 @@ class Provider<T> extends ValueDelegateWidget<T>
...
+  /// ```
+  static void Function<T>(T value) debugCheckInvalidValueType = <T>(T value) {
+    assert(() {
+      if (value is Listenable || value is Stream) {
+        throw FlutterError('''
+Tried to use Provider with a subtype of Listenable/Stream ($T).
+
+This is likely a mistake, as Provider will not automatically update dependents
+when $T is updated. Instead, consider changing Provider for more specific
+implementation that handles the update mecanism, such as:
```
chinitadelrey, line 251: mechanism

PR #117, opened 2019-06-10
Include ProxyProvideron table Existing providers

rubensdemelo:
Include ProxyProvideron table Existing providers


rrousselGit:
Thanks! But I'm not sure if that table is truly helpful.

"Proxy providers" all works the same way, and a dedicated explanation of what they do in general should be better.

rubensdemelo:
Ok, Im gonna try to improve that description.

rrousselGit:
I'll close this since it is already covered from another PR

PR #121, opened 2019-06-12
fix clerical error for Consumer2

Vadaski:
fix a clerical error in the build function passing in the parameter
「chi」 => 「child」


rrousselGit:
Thanks!

That description gave me a heart attack thinking that I'd have to bump major because of it, but thanks! 😄

Vadaski:
You're welcome. 
Almost no effect 😆, I found it when I wrote a guide to the provider.

kevin321happy:
nihao

PR #126, opened 2019-06-20
Displaying a "loading" widget while StreamProvider and FutureProvider work in the background

Alfie-AD:
Hi. Thanks for the hard work. Think this might come in handy. Let me know.!

### DIFF
--- a/packages/provider/lib/src/async_provider.dart
+++ b/packages/provider/lib/src/async_provider.dart
@@ -139,16 +139,17 @@ class StreamProvider<T> extends ValueDelegateWidget<Stream<T>>
       stream: delegate.value,
       initialData: initialData,
       builder: (_, snapshot) {
-        if (snapshot.connectionState == ConnectionState.none) {
-          return Center(child: const Text('No Data'));
-        } else if (snapshot.connectionState == ConnectionState.done) {
-          return InheritedProvider<T>(
-            value: _snapshotToValue(snapshot, context, catchError, this),
-            child: child,
-            updateShouldNotify: updateShouldNotify,
-          );
-        } else if (snapshot.connectionState == ConnectionState.waiting) {
-          return Center(child: const CircularProgressIndicator());
+        if (snapshot.hasError) return Text('Error: ${snapshot.error}');
+
+        switch (snapshot.connectionState) {
+          case ConnectionState.waiting:
+            return Scaffold(body: Center(child: const Text('Loading...')));
+          default:
+            return InheritedProvider<T>(
+              value: _snapshotToValue(snapshot, context, catchError, this),
+              child: child,
+              updateShouldNotify: updateShouldNotify,
+            );
         }
       },
     );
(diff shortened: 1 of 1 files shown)


rrousselGit:
Tanks for the contribution!

On the other hand, such logic should not be performed on the provider, but the consumer. 

StreamProvider is not a StreamBuilder. If your using StreamProvider this way, then you must likely want StreamBuilder instead.

Alfie-AD:
Hi again. i think being able to access your streams deep inside your widget tree without having to pass them in through all those ancestors is the real benefit of Streamprovider, THANKS AGAIN, and that's exactly what brought me here.

Firstly, I get an error the split second before the underlying StreamBuilder of StreamProvider emits any value. That's what got me to contribute. That can't possibly be a good thing in a production app.

This pic is a result of "childCount: streamedList.length" in a SliverChildBuilderDelegate

Screenshot_1561107917

Secondly, i think it might be better to be proactive than reactive. What Consumer exposes to its decendant is the final value which is well past the loading phase of StreamProvider/StreamBuilder and FutureProvider/FutureBuilder. In my case, all i can check is whether the value is empty or not, to the best of my knowledge.

My contribution only passes a widget to the underlying StreamBuiler and FutureBuilder and I hope it's not affecting the logics of Provider package with that.

If am wrong and there indeed is a way to check the status of the streams during the async operations before their final values arrive, kindly let me knw. A sample code will be appreciated.
Thanks.

rrousselGit:
By default, it uses `null` as the loading state.

```dart
StreamProvider<int>(
  builder: (_) async* {
    await Future.delayed(Duration(seconds: 2));
    yield 42;
  },
  child: Consumer<int>(
    builder: (_, value, __) {
      if (value == null) return CircularProgressIndicator();

      return Text('value: $value');
    }
  ),
);
```


This can be tweaked using the `initialData` parameter on `StreamProvider`:

```dart
StreamProvider<int>(
  initialData: 0,
  builder: <something>,
)
```

____


That can be used in combination with "sealed classes" to express your state:

```dart
abstract class MyState {}

class Loading implements MyState {
  const Loading();
}

class Error implements MyState {
  Error(this.error);

  final dynamic error;
}

class Value implements MyState {
   Value(this.count);

   final int count;
}
```

then used this way:

```dart
StreamProvider<MyState>(
  initialData: const Loading(),
  catchError: (_, err) => Error(err),
  builder: (_) async* {
    final result = await doSomeHttpRequest();
    yield Value(result);
  }
  child: Consumer<MyState>(
    builder: (_, state, __) {
      if (state is Loading) return CircularProgressIndicator();
      else if (state is Error) {
        return Text(state.error.toString());
      } else if (state is Value) {
        return Text(state.value.toString());
      }
    }
  ),
);
```

Alfie-AD:
Thanks alot.

rrousselGit:
Does that solve the use case?

If so, can we close this? :smile:

Alfie-AD:
Same error with the first snippet you gave. I guess value is created after streambuilder is done.

I haven't tried the second though, feels cumbersome.
Will let u know how that works out. In the mean time we can close this. 👍

rrousselGit:
> Same error with the first snippet you gave

That's because like I said earlier, in the first snippet, `Provider.of<Foo>` will return `null` when loading.

Simply treat `null` as the loading state 😄

PR #131, opened 2019-06-27
Add state notifier

wesbillman:
This is a concept for adding immutable state handling to provider. I'm not sure if you're intested in a changed like this, but wanted to propose before I put too much effort into making a nice PR :)

Please let me know if you think something like this would be useful and I can clean this up.

Another thought/step would be to add a `StateNotifierProvider` for Widgets to subscribe to state changes rather than the entire bloc `ChangeNotifier`.

```dart
class AuthenticationState extends State {
  final bool isLoading;
  final bool isAuthenticated;

  AuthenticationState({@required this.isLoading, @required this.isAuthenticated})
      : super([isLoading, isAuthenticated]);

  factory AuthenticationState.initial() {
    return AuthenticationState(isLoading: false, isAuthenticated: false);
  }

  factory AuthenticationState.loading() {
    return AuthenticationState(isLoading: true, isAuthenticated: false);
  }
}

class AuthenticationBloc extends StateNotifier<AuthenticationState> {
  AuthenticationBloc({@required this.session});

  logIn(String token) async {
    setState(AuthenticationState.loading());
    // do login stuff...
    setState(AuthenticationState(isLoading: f, isAuthenticated: false));
  }

  logOut() async {
    setState(AuthenticationState.loading());
    // do logout stuff.. 
    setState(AuthenticationState(isLoading: false, isAuthenticated: false));
  }

  @override
  AuthenticationState get initialState => AuthenticationState.initial();
}
```

### DIFF
--- a/packages/provider/pubspec.yaml
+++ b/packages/provider/pubspec.yaml
@@ -12,6 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  equatable:
 
 dev_dependencies:
   pedantic: ^1.4.0
--- a/packages/provider/lib/src/state.dart
+++ b/packages/provider/lib/src/state.dart
@@ -1,0 +1,1 @@
+abstract class State {}
--- a/packages/provider/lib/src/state_notifier.dart
+++ b/packages/provider/lib/src/state_notifier.dart
@@ -1,0 +1,17 @@
+import 'package:flutter/foundation.dart';
+
+abstract class StateNotifier<T> extends ChangeNotifier {
+  T _state;
+
+  StateNotifier() {
+    _state = initialState;
+  }
+
+  T get initialState;
+  T get state => _state;
+
+  void setState(T state) {
+    _state = state;
+    notifyListeners();
+  }
+}


packages/provider/pubspec.yaml
```diff
@@ -12,6 +12,7 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
+  equatable:
```
rrousselGit: Why add this dependency?

rrousselGit:
I looks similar to something I made during my discussions with the flutter_bloc maintainers

https://gist.github.com/rrousselGit/b85ff28f3ee9509a97171f5b944890b1

It's an interesting idea, although a few things:
- that probably should be in an external package (although it can be in this repo)
- I don't like the idea of having `State` be a ChangeNotifier.

Alternatively, it sounds awfully similar to `ValueListenable` too.

packages/provider/pubspec.yaml
wesbillman: Mostly for unit testing states. For example, I can trigger and action, then assert that the states that were emitted match my expectations. This is 100% option, just wanted to through it in here for discussion

```dart
      subject.logIn('token');
      //wait for whatever
      expect(states, [
        AuthenticationState(isLoading: true, isAuthenticated: false),
        AuthenticationState(isLoading: false, isAuthenticated: true),
      ]);
```

wesbillman:
@rrousselGit thanks so much for the quick review and comments.

Yes, this is very similar to the flutter_bloc stuff, and definitely inspired by it. Also, I didn't take this PR too far because I wanted to check in with you first to see if you had and interest and/or if you had a suggestion to solve this in another way (possibly ValueListenable which I'll check out now).

I really like the flutter_bloc concept, but personally prefer the simplicity and usage of provider. I had implemented a project in "pure provider" meaning I didn't have immutable states, but rather just listed to bloc changes and evaluated the properties. For my team, it feels a bit easier to have immutable states to keep us all implementing things in a very standard way with slightly simpler unit test patterns. 

If this is possible/better with `ValueListenable`, perhaps I can give it a go and update the README to show and example of using this package with immutable states (but not really the actions like flutter_bloc/redux).

rrousselGit:
From the earlier gist, we can replace the `Store<T>` class with `ValueNotifier<T>`.

Then we can do:

```dart
class Counter extends ValueNotifier<int> {
  Counter(): super(0);

  void increment() => value++;

  void decrement() => value--;
}
```

and use it this way:

```dart
StoreProvider<int, Counter>(
  builder: (_) => Counter(),
  child: MyApp(),
);
```

And we can then independently do both:

```dart
Provider.of<Counter>(context);
Provider.of<int>(context);
```

The only issue is that `Counter.value` is not flagged as `@protected` or private, so I'm mixed about it.

Here's the updated gist to use `ValueNotifier`  https://gist.github.com/rrousselGit/4910f3125e41600df3c2577e26967c91

wesbillman:
Closing this out after chatting with @rrousselGit. We can pick it up again in the future if it's useful. Thanks @rrousselGit for the review!!

PR #133, opened 2019-06-28
new: make Consumer compatible with MultiProvider

rrousselGit:
Since `Consumer` already takes an optional `child` argument that is passed to `builder`, `Consumer` can be made compatible with `MultiProvider`.

It can be useful for simple scenarios where a `ProxyProvider` doesn't exist but `Provider.of(context, listen: false)` is not possible:

```dart
MultiProvider(
  providers: [
    ChangeNotifierProvider(builder: (_) => Foo()),
    Consumer(
      builder: (_, foo, child) => FutureProvider.value(value: foo.future, child: child),
    ),
  ],
);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,21 @@ Provider<String>.value(
 );
 ```
 
+The widget `Consumer` can also be used inside `MultiProvider`. To do so,
+it must returns the `child` passed to `builder` in the wiget tree it creates.
+
+```dart
+MultiProvider(
+  providers: [
+    Provider(builder: (_) => Foo()),
+    Consumer<Foo>(
+      builder: (context, foo, child) =>
+        Provider.value(value: foo.bar, child: child),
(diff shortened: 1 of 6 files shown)


rrousselGit:
cc @FilledStacks 

If I remember correctly, you needed something

FilledStacks:
@rrousselGit Yes I did. I like getting the Provided value with listen: false. When would listen:false not be available?

rrousselGit:
listen: false won't work if the Future instance can change

rrousselGit:
This can be useful to deal with futures/other.

Instead of listening the future inside ChangeNotifier, the notifier can expose the future itself and bind it to FutureProvider.

FilledStacks:
Noted. I'll keep this in mind and see if any required usage of it comes up while building apps.

PR #149, opened 2019-07-10
Fix code formatting errors

toureh:
`flutter format --set-exit-if-changed lib test` is breaking

https://travis-ci.org/rrousselGit/provider/builds/556749587


rrousselGit:
Thanks!

Issue #161, opened 2019-07-17
Long lines in comments


bradyt:
Thank you for your awesome package.

Please turn on `- lines_longer_than_80_chars` in your `analysis_options.yaml`. It makes is easier to use the same reading environment across projects. Toggling a wrapping view in my editor is not as readable as hard wraps.

PR #162, opened 2019-07-18
Add linting of lines longer than 80 chars

bradyt:
I wasn't entirely consistent everywhere, but this commit is just a sketch of adding the linter.

Would fix https://github.com/rrousselGit/provider/issues/161.

I could also split out the part about uri missing in widget test, and maybe script a search for typos, and add a full commit for that.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -40,6 +40,7 @@ linter:
     - library_names
     - library_prefixes
     - list_remove_unrelated_type
+    - lines_longer_than_80_chars
     - no_adjacent_strings_in_list
     - no_duplicate_case_values
     - non_constant_identifier_names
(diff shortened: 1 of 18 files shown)


packages/provider/example/test/widget_test.dart
```diff
@@ -8,6 +8,7 @@
 import 'package:flutter/material.dart';
 import 'package:flutter_test/flutter_test.dart';
 
+// ignore: uri_does_not_exist
```
rrousselGit: I don't have this issue. Why is it needed?

bradyt: Okay, I removed it and filed as distinct issue, at https://github.com/rrousselGit/provider/issues/164.

rrousselGit:
Thanks!

PR #165, opened 2019-07-20
Check AsyncSnapshot.hasData

tvolkert:
This updates the retrieval of `AsyncSnapshot.data` to first
check `AsyncSnapshot.hasData`, in accordance with the guidance
laid out by https://github.com/flutter/flutter/issues/34545

### DIFF
--- a/packages/provider/lib/src/async_provider.dart
+++ b/packages/provider/lib/src/async_provider.dart
@@ -136,7 +136,7 @@ class StreamProvider<T> extends ValueDelegateWidget<Stream<T>>
   Widget build(BuildContext context) {
     return StreamBuilder<T>(
       stream: delegate.value,
-      initialData: initialData,
+      initialData: initialData, // ignore: deprecated_member_use
       builder: (_, snapshot) {
         return InheritedProvider<T>(
           value: _snapshotToValue(snapshot, context, catchError, this),
@@ -164,7 +164,7 @@ Exception:
 ${snapshot.error}
 ''');
   }
-  return snapshot.data;
+  return snapshot.hasData ? snapshot.data : null;
 }
 
 class _StreamControllerBuilderDelegate<T>
@@ -172,7 +172,7 @@ class _StreamControllerBuilderDelegate<T>
   _StreamControllerBuilderDelegate(this._builder) : assert(_builder != null);
 
   StreamController<T> _controller;
-  ValueBuilder<StreamController<T>> _builder;
+  final ValueBuilder<StreamController<T>> _builder;
 
   @override
   Stream<T> value;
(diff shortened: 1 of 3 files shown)


tvolkert:
@rrousselGit @bradyt @tp please take a look.

This is a forwards-compatible change that should allow users to use the package in both Flutter master and stable.

packages/provider/lib/src/async_provider.dart
```diff
@@ -164,15 +164,15 @@ Exception:
 ${snapshot.error}
 ''');
   }
-  return snapshot.data;
+  return snapshot.hasData ? snapshot.data : null;
```
rrousselGit: There's no reason to check `hasData` here. This is redundant with `initialData`

tvolkert: At tip-of-tree Flutter, the deprecated `initialData` field will trigger a "no data" snapshot if the `initialData` is  null, so the check is still needed.

Once the Flutter change has made it through to the stable channel, you can update this use `.withInitialData()`, and then the `hasData` check won't be needed anymore.

tvolkert: (for this reason, without the changes in this PR, the tests fail when run against Flutter with https://github.com/flutter/flutter/pull/36577 applied)

rrousselGit: But on master, the. `provider` would use `StreamBuilder.withInitialData`.

I'd prefer to stop using `StreamBuilder` entirely if that's an issue. `provider` voluntarily don't expose `AsyncSnashot` to begin with.

tvolkert: I see.  So the other way to go would be https://github.com/rrousselGit/provider/pull/166, which puts an upper bound on the Flutter SDK to prepare for  the breaking change.  Then this would be updated to use the new API instead  of the deprecated one (and would be updated to version `3.1.2` and would contain a _lower bound_ Flutter SDK constraint).

wdyt?

rrousselGit: Hum I don't like that either.
Using StreamBuilder here is just an implementation details. 

The best alternative is for `provider` to stop using `StreamBuilder`, and do the listening itself.
That should even give a small performance boost, considering `provider` don't need `AsyncSnashot` at all.

tvolkert: Ok.  I'm not familiar with the implementation of `provider` -- are you game to update it to not use `StreamBuilder`?

rrousselGit: Sure. I'll do that.. Although depending on where https://github.com/flutter/flutter/issues/34545#issuecomment-513482773 is going, there may be no need for a change at all.

rrousselGit:
@tvolkert should this be closed since the change has been reverted?

tvolkert:
No, I'm still gonna reapply the smaller change of making the data accessor throw if there's an error

rrousselGit:
But that specific change is not breaking for `StreamProvider`. It already checks `hasError` and reports it to the `catchError` callback or `FlutterError.reportError` if not callback is provided.

tvolkert:
Oh sorry - I thought I was commenting on flutter/flutter#34545.  Yes, this can be closed 😄

Issue #172, opened 2019-07-24
Explain Consumer performance optimizations. 


feinstein:
The README.md docs say `Consumer` can be used for performance optimizations, but don't explain what are these optimizations. I think it refers to using the `child` property, but I only learned about it in other tutorials. I think the docs should be improved from a beginner's perspective.

anticafe:
In fact I still don't understand what's the different between using `Consumer` vs. without `Consumer`.

feinstein:
As far as I understood, `Consumer` can provide a `child` along with the value. This child is a `Widget` tree that you made before and gave it to `Consumer`. This ʼWidget` tree is sopposed to be immutable, so you only create it once, and reuse it, so if the value is updated only the `Widgets` that need to be updated are reconstructed.

I think an example will be more clear that my explanation, hence the request for better documentations. 

Also, there may be other differences that I am not aware of.

rrousselGit:
Ah, actually I explained that in a recent talk.
I'll add it to the documentation soon.

TL;DR: Consumer vs Provider.of changes the widget that rebuilds. 

With Provider.of, it's the widget that called that method which rebuilds. With Consumer, it's the Consumer. 

That combined with the optional `child` allows for more gralunar rebuild of the widget tree.

feinstein:
When you say `With Provider.of, it's the widget that called that method which rebuilds`, you mean that the `Widget` that we are coding the `build` method, the one that we got `context` from, will be rebuild, right? And not the `Widget` that we placed `Provider.of` inside, like inside a `RaisedButton` `onPressed` method in the middle of the `build` method.

rrousselGit:
> you mean that the Widget that we are coding the build method, the one that we got context from, will be rebuild, right?

Yes. It's the widget associated with the `BuildContext` passed to `Provider.of` that will rebuild when the value changes.

We could technically use `Builder` instead of `Consumer`. It'd do the same thing.

feinstein:
It will be nice to have an example at the docs, showing who will get rebuild, as the technical description can be ambiguous for begginers.

rrousselGit:
https://github.com/rrousselGit/provider/pull/176

I opened a PR that improves the documentation on `Consumer`. Is that enough?

rrousselGit:
> Also, does it speaks about dispose?

What dispose?

feinstein:
The dispose method for when the provider goes away and we can destroy our resources.

feinstein:
[...]

PR #176, opened 2019-07-26
Documentation

rrousselGit:
closes https://github.com/rrousselGit/provider/issues/172


feinstein:
"But `Foo` changes, then both `BarWidget` _and_ `FooWidget` will rebuild."

Wouldn't it be "But when Foo changes"?

Also, there's no Foo in the diff, maybe it's on the original code, but I can't see it right now.

rrousselGit:
Thanks. Fixed

feinstein:
Thanks great.

Issue #182, opened 2019-07-29
Testing widget which are using provider


rubiktubik:
Hi, i'am using the great provider package in my project.

I was about to write a widget test. And there is no documentation how to that with
provider. 

I have my widget:
```dart
class MyWidget extends StatelessWidget {
  @override
  Widget build(BuildContext context) {
    return Consumer<MyViewModel>(
      builder: (BuildContext context, MyViewModel myVm,
          Widget child) { ... }
```

And i have my definition of providers:
```dart
runApp(
    MultiProvider(
      providers: <SingleChildCloneableWidget>[
        ChangeNotifierProvider<MyViewModel>(
            builder: (BuildContext context) => MyViewModel()),
      ],
      child: MyApp(),
    ),
  );
```

And i have my test:

```dart
void main() {
  testWidgets('Should work as expected',
      (WidgetTester tester) async {
    await tester.pumpWidget(MyWidget());
  });
}
```

How can i test my widget, with using a mock viewmodel?

Regards 
Michael

rrousselGit:
Hello!

You need to add the necessary providers in the widget tree passed to `pumpWidget`:

```dart
await tester.pumpWidget(
  Provider.value<MyViewModel>(
    value: someMock,
    child: MyWidget(), 
  ),
) 
```

rubiktubik:
Thank you for the quick answer! 
I'am using mockito to mock my ViewModel but i get 
```dart
The following ProviderNotFoundError was thrown building Consumer<MyViewModel>(dirty):
Error: Could not find the correct Provider<MyViewModel> above this
Consumer<MyViewModel> Widget
```

In the pumpWidgetCode i only changed Provider to ChangeNotifierProvider:
```dart
class MockMyViewModel extends Mock implements MyViewModel{}
...
Widget createWidgetForTesting({Widget child}) {
    return MaterialApp(
      home: child,
    );
  }
...
final mockVm = MockMyViewModel();
await tester.pumpWidget(ChangeNotifierProvider.value(
      value: mockVm,
      child: createWidgetForTesting(child: MyWidget()),
    ));
```

The **createWidgetForTesting** is for using MediaQuery in my widget.

How can i make i work?

rrousselGit:
You need to manually specify the object type when passed to provider:

DON'T :

```dart
Provider.value(
  value: FooMock() 
)
```

DO:

```dart
Provider<Foo>.value(
  value: FooMock(),
) 
```

[...]

Issue #170, opened 2019-07-23
Make it harder to use builder constructors with an existing object


rrousselGit:
Folks seems to use `ChangeNotifierProvider()` by having `builder` return an existing `ChangeNotifier` instead of creating a new one:

```dart
var notifier = ValueNotifier(0);

ChangeNotifierProvider(
  builder: (_) => notifier,
  child: ...,
)
```

This is bad, and instead `ChangeNotifier.value` should be used:

```dart
var notifier = ValueNotifier(0);

ChangeNotifierProvider.value(
  value: notifier,
  child: ...
)
```

____

There are a few things we can do:

- document that behavior (on the readme &/or or an article)
- add an assert in the default constructor, that fails if the returned `ChangeNotifier` already has listeners.
- Add an assert in `ChangeNotifierProxyProvider` if the `ChangeNotifier` instance ever changes, as it's unlikely to be desired.

Issue #142, opened 2019-07-05
ProxyProvider6


timrijckaert:
Hi, 

Currently the `ProxyProvider` class has a maximum of 6 type parameters.
However I have a case where I would like more.

Could more be added, possibly even be code generated to a more sane amount?

```
ProxyProvider6<Dio, UserStorage, VRTThemeBuilder, VRTNavigator, DeepLinkHelper, SomeRepository, AppBloc>(
  dispose: (context, val) => val.dispose(),
  builder: (
    BuildContext context,
    Dio dio,
    UserStorage userStorage,
    VRTThemeBuilder vrtThemeBuilder,
    VRTNavigator vrtNavigator,
    DeepLinkHelper deepLinkHelper,
    SomeRepository repository,
    AppBloc previous,
  ) =>
      AppBloc(
        tabBarChoices.first,
        dio,
        _defaultEnv,
        userStorage,
        vrtThemeBuilder,
        vrtNavigator,
        repository,
        deepLinkHelper,
      ),
```

Thanks in advance

Tim

rrousselGit:
How many do you need?

rrousselGit:
Note that you can call `Provider.of(context)` inside `builder` for an unlimited amount.

That'll work

timrijckaert:
Currently I need 7.
I could make it work using the `Provider.of(context)` _trick_, but it kinda feels like a limitation in the `ProxyProvider` API.

Instead it would be nice if you/we could somehow generate a `ProxyProvider` with an arbitrary amount of type parameters.

What do you think?

rrousselGit:
> I could make it work using the Provider.of(context) trick, but it kinda feels like a limitation in the ProxyProvider API.

That's a design choice. The generic argument is just there to make it easier to use, but it works with any inherited widgets.

timrijckaert:
Hmm, so I guess this can be closed then?

rrousselGit:
Hum no. I'll leave this open as a reminder to document that aspect.

And if peoples want some with more arguments, they can :+1: or make a PR. If its often requested, I'll make some more.

MrMonotone:
I ran into this issue as well. I think that it would be good to not only make note of it in the documentation, but include an example using the Provider.of(context) trick as well. I think that would help people out.

Issue #136, opened 2019-07-02
Recover StreamProvider documentation?


junjizhi:
I found a few lines of explanations of StreamProvider in this PR: https://github.com/rrousselGit/provider/pull/21/files 
but the current README doesn't have these explanations anymore. Any reason why this doc is removed?

As far as I understand, StreamProvider is a wrapper of StreamBuilder. It can only listen to streams and consume its values. If we have a widget that both listens to the stream and changes the values of the stream, it's more straightforward to use `ChangeNotifierProvider`. Is that right?

I was trying to see how StreamProvider works, but because of too little documentation, I had to spend two hours to dig out everything I could find. While this is an amazing package, more doc will definitely help.

Thanks!

PR #185, opened 2019-08-01
Usage documentation

### LINKED ISSUE #139: Only the first instance returned by builder of ProxyProvider (ChangeNotifierProxyProvider) is propagated down the tree
Expected behaviour:
1. Every time the floating action button is pushed, `DecCounter`'s decimal value is incremented.
2. `DecCounter` extends `ValueNotifier`, so the change of value causes the builder of `ListenableProxyProvider<DecCounter, HexCounter>()` to be called.
3. In the builder, `HexCounter`, which also extends `ValueNotifier`, is newly instantiated with the decimal value passed in.
4. The value is converted to a hexadecimal number in `HexCounter`.
5. The change is notified and an update of Text should be triggered.

```dart
class DecCounter extends ValueNotifier<int> {
  DecCounter() : super(0);

  void increment() => value++;
}

class HexCounter extends ValueNotifier<String> {
  HexCounter(int decimal) : super(decimal.toRadixString(16));
}
```

```dart
import 'package:provider/provider.dart';

void main() => runApp(MyApp());

  @override
    return MaterialApp(
      home: MultiProvider(
        providers: [
          ListenableProvider<DecCounter>(
            builder: (_) => DecCounter(),
            dispose: (_, decCounter) => decCounter.dispose(),
          ),
          ListenableProxyProvider<DecCounter, HexCounter>(
            builder: (_, decCounter, __) => HexCounter(decCounter.value),
            dispose: (_, hexCounter) => hexCounter.dispose(),
          ),
        ],
        child: Scaffold(
          body: _CounterView(),
          floatingActionButton: Consumer<DecCounter>(
            builder: (_, counter, __) {
              return FloatingActionButton(
                onPressed: counter.increment,
                child: const Icon(Icons.add),
              );
            },
          ),
        ),
      ),
    );
  }
}

class _CounterView extends StatelessWidget {
  @override
    final counter = Provider.of<HexCounter>(context);

    return Center(
      child: Text(counter.value.toString()),
    );
  }
}
```

This code does not work as I expected. The number on the screen is not incremented.

I tried many changes in my code and managed to get it to work finally by using and returning the existing instance of `HexCounter` that is obtained from the last parameter of the builder.

```diff
 class HexCounter extends ValueNotifier<String> {
-  HexCounter(int decimal) : super(decimal.toRadixString(16));
+  HexCounter() : super('0');
+
+  void setValue(int decimal) {
+    value = decimal.toRadixString(16);
+  }
 }
```

```diff
 ListenableProxyProvider<DecCounter, HexCounter>(
-  builder: (_, decCounter, __) => HexCounter(decCounter.value),
+  initialBuilder: (_) => HexCounter(),
+  builder: (_, decCounter, prevHexCounter) {
+    return prevHexCounter..setValue(decCounter.value);
+  },
   dispose: (_, hexCounter) => hexCounter.dispose(),
 ),
```

It looks like returning different `HexCounter` instances by the builder stops propagation. Am I right in thinking so? Does the builder of `ProxyProvider` / `ChangeNotifierProxyProvider` / `ListenableProxyProvider` have to return the same instance on every rebuild?

If so, it seems the documentation of `ProxyProvider`'s constructor may need a little more detail to prevent misuse.

> As opposed to the builder parameter of Provider, builder may be called more than once.

This misled me to think the builder was allowed to return a new instance every time.

### LINKED ISSUE #171: Improve usage documentation
I am starting to learn about `provider` and it's hard to me to understand the differences and use cases of many of the different providers inside the table at the end of the README.md file. For example:

1. `ListenableProvider` is supposed to listens to a `Listenable`, but as fart as I can see, `Listenable` is just `an object that maintains a list of listeners`, it doesn't trigger any updates, so how will `ListenableProvider`  `ask widgets which depend on it to rebuild whenever the listener is called`, if there are no notification mechanisms in `Listenable`? Same question could be asked for `ValueListenableProvider`.

2. What's the difference between `ValueListenableProvider.value`, `ListenableProvider.value` and `ChangeNotifierProvider.value`?

3. For the sake of completeness, shouldn't `ProxyProvider` and `MultiProvider` also be at the table?

Simple examples, along with more refined documentations, showing those use cases would be more than welcome.

rrousselGit:
fixes #139
closes #171 #170 #142 #182 #136

### DIFF
--- a/packages/provider/CHANGELOG.md
+++ b/packages/provider/CHANGELOG.md
@@ -1,5 +1,8 @@
 # 3.1.0
 
+- improved the overall documentation
+- fixed a bug where `ChangeNotifierProvider.value` didn't update dependents
+  when the `ChangeNotifier` instance changed.
 - `Consumer` can now be used inside `MultiProvider`
   ```dart
   MultiProvider(
--- a/packages/provider/lib/src/async_provider.dart
+++ b/packages/provider/lib/src/async_provider.dart
@@ -17,6 +17,13 @@ typedef ErrorBuilder<T> = T Function(BuildContext context, Object error);
 
 /// Listens to a [Stream<T>] and exposes [T] to its descendants.
 ///
+/// Its main use-case it to provide to a large number of a widget the content
+/// of an existing stream, without caring about reacting to events.
+///
+/// A typical example would be to expose the battery level, or a Firebase query.
+/// Trying to use [Stream] to replace [ChangeNotifier] is outside of the scope
+/// of this class.
+///
 /// It is considered an error to pass a stream that can emit errors without
 /// providing a [catchError] method.
 ///
--- a/packages/provider/lib/src/change_notifier_provider.dart
+++ b/packages/provider/lib/src/change_notifier_provider.dart
@@ -18,7 +18,8 @@ import 'proxy_provider.dart';
 /// leaks and potentially undesired side-effects.
 ///
 /// See [this stackoverflow answer](https://stackoverflow.com/questions/52249578/how-to-deal-with-unwanted-widget-build)
-/// which explains in further details why this is undesired.
+/// which explains in further details why using the `.value` constructor to
+/// create values is undesired.
 ///
 /// - DO create a new [ChangeNotifier] inside `builder`.
 /// ```dart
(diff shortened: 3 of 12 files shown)


feinstein:
I found some English mistakes on the docs, I can't review all the code now, but I will make a list later.

feinstein:
@rrousselGit did you see this?

rrousselGit:
Yes, thanks. I fixed them (although some of these are not mistakes, like the `..foo = foo`)

feinstein:
Ok, I thought if `foo` was an instance variable `.` would be enough.

PR #198, opened 2019-08-21
Mark constructors on @immutable classes as const

kjeremy:
Most of the provided classes are immutable so we can mark the constructors as `const`


rrousselGit:
This is voluntary.
None of these classes can be used with `const` even if the constructor allowed it.
So it doesn't make sense to allow it, nor does it make sense to enable the linter rule.

What is the use-case?

kjeremy:
I noticed later that I couldn't use them in a const context.

PR #210, opened 2019-09-03
add StreamProvider.of

johnxguo:
add StreamProvider.of, to expose the stream to all of [StreamProvider] descendants


rrousselGit:
Hi, thanks for the contribution!

But what is the use-case?

johnxguo:
I found that if i new a stream in the builder of StreamProvider, there's no way to get it from the descendants, is there another way?

rrousselGit:
Why do you want to get the stream to begin with?

PR #243, opened 2019-10-13
V4.0.0

### LINKED ISSUE #81: Lazy load providers
One thing I don't get it:
An app with master and detail views. 
I want to supply a DAO class for each view.
Should I use a Multiprovider right after main() or should I use individual Providers in the views class?

I want to instatiate the classes only when the view is loaded. My question is do they get instantiated already if I put it on the top of the app or do they have like lazy instantiation

_Originally posted by @peekpt in https://github.com/rrousselGit/provider/issues/77#issuecomment-494551544_

### LINKED ISSUE #213: Consider adding diagnostic children to Providers
Being able to see a Provider's current value and the value's identityHashCode in the dev tools would be super useful.

### LINKED ISSUE #236: Make the flag "listen" of "Provider.of" default to "false" if called when the widget tree is not building
### Problem:

It is too easy to forget to specify `listen: false` on a click handler / ...
This causes undesired effects, where after an event handler, the handler widget suddenly rebuilds too often.

### Solution:

Since event handlers are **always** called when the widget tree isn't building, we can automatically infer what `listen` should be

### Consequence:

```dart
RaisedButton(
  onPressed: () {
    Provider.of<Whatever>(context); 
  }
)
```
^ This code will now break.
Before, it caused the widget that creates `RaisedButton` to rebuild when `Whatever` changes.

Now, no rebuild will be caused.

For production code, this is usually not an issue.
Where this has an impact, on the other hand, is for tests.

If you previously called `Provider.of` after a `tester.pumpWidgets` to test if something was rebuilding, then your test may now fail.

### LINKED ISSUE #237: Simplify the integration of custom widgets with MultiProvider
### Problem:

`MultiProvider` is a very important feature for a provider. But it is currently tedious and error-prone to make a widget compatible with `MultiProvider`:

- we have to implement `SingleChildCloneableWidget`, that is very error-prone (easy to forget a parameter)
- it usually requires an extra private constructor, just for the clone implementation

All in all, it's too complex, which makes peoples not want to make custom providers and stick to the existing ones.

### Solution

I made a POC of an alternate implementation of MultiProvider, which doesn't relly on a "clone" method.

Instead, it uses a new kind of widget (neither Stateless nor Stateful): SingleChildStatelessWidget / SingleChildStatefulWidget.

The difference is that the `build` method takes an extra `Widget child` parameter:

```dart
class Example extends SingleChildStatelessWidget {
  Example({Key key, Widget child}): super(key: key, child: child);

  @override
  Widget build(BuildContext context, {Widget child}) {
    // TODO: do something with `child`
  }
}
```

Which is then instantly compatible with `MultiProvider`:

```dart
MultiProvider(
  providers: [
    Example(),
  ],
)
```

### Consequences:

The `SingleChildCloneableWidget` interface will be removed, and replaced by a `SingleChildWidget` interface and two implementations: 
- `SingleChildStatelessWidget`
- `SingleChildStatefulWidget`

`MultiProvider` will then accept a `SingleChildWidget` instead of `SingleChildCloneableWidget`

Existing implementations of `SingleChildCloneableWidget` should be migrated to use one of the two `SingleChildWidget` implementation instead.

Before:

```dart
class Example extends StatelessWidget implements SingleChildCloneableWidget {
  Example({Key key, int value, this.child}): this(key: key, foo: Foo(value), child: child);

  Example._({Key key, this.foo, this.child): super(key: key);

  final Foo foo;
  final Widget child;

  @override
    return Whatever(
      foo: foo,
      child: child,
    );
  }

  @override
  Example cloneWithChild(Widget child) {
    return Example._(
      key: key,
      foo: foo,
      child: child,
    );
  }
}
```

After:

```dart
class Example extends SingleChildStatelessWidget {
  Example({Key key, int value, Widget child}): foo = Foo(value), super(key: key, child: child);

  final Foo foo;

  @override
  Widget build(BuildContext context, {Widget child}) {
    return Whatever(
      foo: foo,
      child: child,
    );
  }
}
```

### LINKED ISSUE #259: [RFC] Renaming the parameter 'builder' of providers
Hello!

TD;DR:
This proposal is for a breaking change: renaming `builder` of `Provider`/`ChangeNotifierProvider`/..., to either `initialValue` or `initialValueBuilder`.

Which means before:

```dart
Provider<int>(
  builder: (context) => 42,
);
```

after:

```dart
Provider<int>(
  initialValue: (context) => 42,
);
```

or `initialValue` -> `initialValueBuilder`.

*Feel free to suggest other names or express any point of view on the topic.*

### Motivation

Currently,  the default constructor of providers all has a named required parameter `builder`, that builds a value _once_.

But this name causes a few issues.

- it's misleading.
  In the context of Flutter, a parameter named `builder` is usually a function that returns a _widget_, not a model. Such as:

  ```dart
  StreamBuilder(
    builder: ...
  )
  ```

  Renaming it to will remove the confusion.
  And a name such as `initialValue` also makes it more apparent that `builder` will *not* be called again.

- it prevents fusing `*Provider` and `*ProxyProvider`.
  These two widgets are _very_ similar.
  Usually, `*Provider` is just syntax sugar for `*ProxyProvider ` with only an `initialBuilder` like so:
  ```dart
  Provider(
    builder: (_) => 42,
  );
  ```
  is equivalent to:
  ```dart
  ProxyProvider<*, int>(
    initialBuilder: (_) => 42,
    builder: (_, __, value) => value,
  );
  ```
  But it is currently impossible to merge both these widgets because they have both a parameter named `builder` but with a different prototype.

  By renaming `builder` to something more expressive, it is possible to host both `*Provider` and `*Proxy` under the same class.

  We could have:
  ```dart
  ChangeNotifierProvider(
    initialValue: (_) => MyNotifier(),
  );
  ```
  Then we'd remove `ProxyProvider`.
  And add an extra `value` or `valueBuilder` parameter on `Provider`:
  ```dart
  ChangeNotifierProvider(
    initialValue: (_) => MyNotifier(),
    value: (context, notifier) {
      // Works like *ProxyProvider.builder
      // we can safely call `Provider.of` here
      return notifier
        ..foo = Provider.of<Foo>(context)
    },
  );
  ```
  Of course, we'd still have numeric syntax sugar like `ProxyProvider2`/..., but without the `Proxy` keyword:
  ```
  Provider1<Dependency, Result>(
    value: (context, Dependency dep, Result previous) {
      return Result(dep);
    },
  )
  ```

- it prevents using `builder` for an _actual_ "builder" that behaves likes Consumer/StreamBuilder.

  For example, a common use-case is to do:

  ```dart
  Provider<T>(
    builder: (_) => T(),
    child: Consumer<T>(
      builder: (_, value, __)  {
        // TODO: use `value`
      },
    ),
  )
  ```
  But that's pretty verbose.

  We could simplify it to:
  ```dart
  Provider<T>(
    initialValue: (_) => T(),
    builder: (_, value, __) {
      // TODO: use `value`
    },
  )
  ```
  Which would be **strictly** equivalent to the previous code – just  smoother to write.
  But such simplification is not possible because `builder` is already taken.

### Transition

This renaming could be done with a smooth transition using `@deprecated`.
In the v3.x, both `builder` and the new name could be on the widget.
`builder` would then be marked as `@deprecated`.

On the other hand, the other listed changes (fusing `*Provider` & `*ProxyProvider` or the Consumer shorthand) would _not_ be part of the v3 but instead the v4 – which will also include loading.

 
Thoughts?

Feel free to comment 👍 if you agree or comment if you have other propositions or disagree with the suggestion.

### LINKED ISSUE #276: Rename ProviderNotFoundError to ProviderNotFoundException
related to #235

rrousselGit:
closes #237, closes #236, closes #259, closes #81, closes #276, closes #213


packages/provider/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
+# 4.0.0
+
+- renamed `ProviderNotFoundError` to `ProviderNotFoundException`.
+  This allows calling `Provider.of` inside a `try/catch` without triggering a
+  warning.
+- update provider to work with Flutter 1.12.1
+- The creation and listening of objects using providers is now performed lazily.
+  This means that objects are created the first time the value is read instead of
```
isacjunior: Nice change 🥰

packages/provider/CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+  the first time the provider is mounted.
```
isacjunior: Nice change 🥰

packages/provider/lib/provider.dart
```diff
@@ -1,12 +1,17 @@
 library provider;
 
+export 'package:nested/nested.dart'
```
felangel: As discussed on Slack can you change this to:

```dart
export 'package:nested/nested.dart' hide Nested;
```

PR #290, opened 2019-12-18
Added second alternative to use async function in init state

isacjunior:
This pull request incresent more informations in read about how to call async function in `initState` with Provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -300,16 +300,26 @@ entire tree equally:
   ```
   This is useful when there's no "external parameter".
 
-- asynchronously at the end of the frame:
+- asynchronously at the end of the frame. Alternative 1:
   ```dart
   initState() {
     super.initState();
     Future.microtask(() =>
-      Provider.of<Foo>(context).fetchSomething(someValue);
+      Provider.of<Foo>(context, listen: false).fetchSomething(someValue);
     );
   }
   ```
   It is slightly less ideal, but allows passing parameters to the mutation.
+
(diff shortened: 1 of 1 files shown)


isacjunior:
After v4.0, information about `listen: false` should be removed.

README.md
```diff
@@ -300,17 +300,27 @@ entire tree equally:
...
   ```dart
   initState() {
     super.initState();
     Future.microtask(() =>
-      Provider.of<Foo>(context).fetchSomething(someValue);
+      Provider.of<Foo>(context, listen: false).fetchSomething(someValue);
     );
   }
   ```
   It is slightly less ideal, but allows passing parameters to the mutation.
```
rrousselGit: Instead of another `-`, what about a "Alternatively, instead of `Future.microtask`, you can use [addPostFrameCallback](https://api.flutter.dev/flutter/scheduler/SchedulerBinding/addPostFrameCallback.html)"


I'm not sure that a complete example is necessary. To begin with, this is a workaround and should be avoided as much as possible.

So I'm not too fan of this approach taking more space on the readme

isacjunior: Ok, I will close this PR, but why `addPostFrameCallBack` should be avoided and what difference between `Future.microtask`?

rrousselGit: It's not that `addPostFrameCallback` should be avoided, but the entire idea of "delay changes at the end of the frame to bypass the initState limitation"

PR #296, opened 2019-12-21
add scope to providers

rrousselGit:


rrousselGit:
Closing for now. It if interesting, but probably too complex and not useful enough to be worth it

rrousselGit:
The branch will not be deleted any time soon though, if you want to fork it or have arguments in its favor

PR #319, opened 2020-01-15
Add "builder" parameter to providers and BuildContext extensions

### LINKED ISSUE #251: Add extension on BuildContext
Hey guys, playing with extensions of dart 2.6, I was thinking on a nice one:

```dart
extension ProviderExtensions<T> on BuildContext {
    T provide<T>() => Provider.of<T>(this);
}
``` 

Then you could use it like:

```dart
  var person = context.provide<Person>();
```

Would be a nice addition after the extensions functions hit stable :)
What you think?

### LINKED ISSUE #334: The getter 'owner' was called on null.
I understand this is more of a question than it is an issue, but I don't know where else to ask.

I am using provider version 4.0.2 and I get this error i the console: "The getter 'owner' was called on null."
I do not know how to resolve this issue. Any help is appreciated! Thank you in advance!

rrousselGit:
closes #251, closes #334

- [x] mention the new features on CHANGELOG.md
- [x] document the new feature inside the README
- [x] add `builder` to providers
- [x] add `context.read`
- [x] add `context.watch`
- [x] add `context.select`
- [x] prevent `context.read` from being used **inside** `didChangeDependencies` and `build`
- [x] prevent `context.watch` from being used **outside** of `didChangeDependencies` and `build`
- [x] add assertions that `select`  can only be called inside `build` (but **not** `didChangeDependencies`)

The assertions currently cannot be implemented without https://docs.google.com/document/d/1WowGJhAMzYvwoyYLOzwjj1Jb0lwIjUpNfqvKjFWQZIM/edit#


smiLLe:
I was trying the new extension but i ran into an issue.
Nested Selectors are working:
```dart
return Selector<GlobalStore, bool>(
      selector: (_, store) => store.state.someProp == null,
      builder: (_, hasSomeProp, __) {
        return Selector<GlobalStore, bool>(
          selector: (_, store) => store.state.someOtherProp,
          builder: (_, isSomeOtherProp, __) {
... do stuff
          },
        );
      },
    );
```

If i replace the Selector code above with this snipped, i will receive an error
``` dart
final isSomeOtherProp = context
        .select<GlobalStore, bool>((store) => store.state.someOtherProp);

final hasSomeProp = context.select<GlobalStore, int>(
            (store) => store.state.someProp) ==
        null;

... do stuff
```

> Cannot call context.read/watch/select inside the callback of a context.select
'package:provider/src/provider.dart':
Failed assertion: line 214 pos 12: '_debugIsSelecting == false'

rrousselGit:
I would need more context. I don't see any issue with just that

rrousselGit:
It's worth noting that this assert is associated with an error message:

```
Cannot call context.read/watch/select inside the callback of a context.select
```


You likely did something similar to:


```dart
context.select((Whatever o) {
  context.read<SomethingElse>();
});
```

smiLLe:
I shared the code i am using. the only thing i changed was the property names to make it easier to understand.
Let me try to reproduce it on a new flutter project.

rrousselGit:
That would be great thanks!

Because calling `select` multiple times is definitely supported. So either its either a bug on your hand or on my hand 🙃

smiLLe:
ok i found it.
i have a ```ChangeNotifierProvider``` like this:
``` dart
ChangeNotifierProvider<Dep, Store>(
create: (context) =>Store(Provider.of<Dep>(context, listen: false)),
updated: (_ dep, __) => Store(dep),
)
```
As soon as is add ```lazy: true```it will work

smiLLe:
I should probably not create my Store like this :) Sorry for invading this PR :P

rrousselGit:
Oh no that's an interesting case that I didn't think about. Thanks!
I'll write a few tests for it

smiLLe:
You consider this a bug? If so, is it fine to call ```Provider.of<T>(c, listen: false)``` in create() ?

rrousselGit:
Yes, it is fine 😄

rrousselGit:
Alright, I've just pushed a fix for it.

Thanks for the report!

smiLLe:
The PR you made will delay this PR for several weeks at least, right :( ?

rrousselGit:
It really depends on how the discussion with the team ends up

If they say no (which they currently seems to want do to), then I'll just release it

rrousselGit:
I can make a dev release first though.

rrousselGit:
Alright, I made a 4.1.0-dev release just in case. It doesn't hurt anybody

hsul4n:
> Alright, I made a 4.1.0-dev release just in case. It doesn't hurt anybody

When will be the official release for the new great API with the power of extensions.

rrousselGit:
When Flutter makes a stable release

PR #321, opened 2020-01-17
Remove description of listen inference

kaboc:
The description of `listen` inference in `Provider.of()` should no longer be necessary after the removal of the feature in v4.0.0-hotfix.1. If I'm wrong, just close this PR please.


rrousselGit:
Oh, nice catch.

Thanks!

PR #346, opened 2020-02-12
Remove indication to always use package imports

### LINKED ISSUE #345: "Always use package imports."
You plugin says to [always use package imports](https://github.com/rrousselGit/provider/blob/77b6cf6599e6fe2179858a6947e36df8b3621550/packages/provider/lib/src/provider.dart#L330), whereas [Dart guidelines](https://dart.dev/guides/language/effective-dart/usage#prefer-relative-paths-when-importing-libraries-within-your-own-packages-lib-directory) mention to prefer relative paths when importing librairies within our own package’s lib directory.

Why would it be better to use package imports when using Provider ? 

I personally always use relative paths imports and have never experienced any problem with Provider.

Thanks

[Stackoverflow question](https://stackoverflow.com/questions/59693195/flutter-imports-relative-path-or-package)

augustinreille:
As mentioned in [this stackoverflow answer](https://stackoverflow.com/a/60184880/10122791), the indication to always use package imports is now obsolete as the mentioned bug from Dart does not exist anymore.

Closes #345

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -309,7 +309,6 @@ To fix, please:
   * Provide types to Provider<$valueType>
   * Provide types to Consumer<$valueType>
   * Provide types to Provider.of<$valueType>()
-  * Always use package imports. Ex: `import 'package:my_app/my_code.dart';
   * Ensure the correct `context` is being used.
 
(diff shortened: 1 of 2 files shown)


rrousselGit:
Perfect, thanks!

PR #348, opened 2020-02-14
Fix misstake in the README.md

woprandi:
Fix the confusion between `create` and `update` for `ProxyProvider` in the doc


rrousselGit:
Indeed thanks!

PR #427, opened 2020-04-24
Add Portuguese translation and base for others

robsonsilv4:
Based on the latest report shared by Flutter's team, Brazil is one of the five places where Flutter is most used.

Brazil different from other places, does not have English as a first or second language. Because of that I think it's important to have a translation to Portuguese

I also prepared a structure for other translations, if needed.


robsonsilv4:
@rrousselGit I updated the pr with justification.
And congrats, Provider it's also one of most packages used.

rrousselGit:
Hey~
Sorry for the delay

Thanks a lot!

PR #446, opened 2020-05-10
Updates the comment doc  of the `ProxyProvider`

futurechallenger:
In the comment, the `return update(context, a, b, result)` for `ProxyProvider<A, Result>` has a extra parameter, which is the `b`. It's  should not appear here.


rrousselGit:
Thanks!

PR #460, opened 2020-05-24
Add FAQ section for consuming provider interface

### LINKED ISSUE #234: Consumer<Interface>
How would one go about consuming an interface / abstract class / mixin? Is this something that's supported. I had suspected that since I am already providing the concrete implementation I could consume the interface, but that results in an exception that the type can't be found.

I had also thought I could use a proxy provider, but my concrete type extends change notifier and my interface doesn't, meaning I can't construct it with ChangeNotifierProxyProvider nor with the standard ProxyProvider.

Thoughts?

KieranCTaylor:
It took me a long time to find the answer for this and saw a few other people struggling along the way too so thought it could be helpful especially as this is a key part of testing with mocks.

closes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -613,6 +613,34 @@ Provider<Country>(
 ),
 ```
 
+#### Can I consume an interface and provide an implementation?
+
+Yes, a type hint must be given to the compiler to indicate the interface will be consumed, with the implementation provided in create.
+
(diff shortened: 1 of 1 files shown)


rrousselGit:
Nice, thanks!

PR #482, opened 2020-07-01
Add builder parameter to MultiProvider

joaomarcos96:
The `builder` parameter allows the immediate use of `context.watch` to consume `MultiProvider` children.

It has the same behavior as the `builder` parameter of a single `Provider`.


rrousselGit:
Nice thanks!

PR #561, opened 2020-10-27
Fix comment

takeru:
Fix comment bug in example/lib/main.dart

```
/// Calls `context.watch` to make [MyHomePage] rebuild when [Counter] changes.
```

to

```
/// Calls `context.watch` to make [Count] rebuild when [Counter] changes.
```


rrousselGit:
Thanks!

Issue #564, opened 2020-10-30
Restoration Framework


nikitadol:
Are there any plans to implement this for providers?
https://docs.google.com/document/d/1KIiq5CdqnSXxQXbZIDy2Ukc-JHFyLak1JR8e2cm3eO4/edit#

rrousselGit:
What are you expecting from provider here?

provider is for business state, which is in the "domain state" category. Domain state if out of the scope of the restoration API

nikitadol:
For example, I have a search provider.
I do not need to store this data between application launches, but I would like the state to persist if the system kills the Activity.

Therefore, it would be convenient if the provider provided an API so that my provider class could save its state

vgribok:
This is not a bug, for sure, and the issue is broader than what Provider can or cannot do, and could be restated as follows: 
> Any framework or library (like Provider or Flutter 2 Navigator) enabling the app to use StatelessWidget+ChangeNotifier pattern, thus making StatefulWidget largely unnecessary, leads to RestorationMixin - which seems to be the cornerstone of the state restoration - being useless due to its dependency on the StatefulWidget. 

To solve this problem, an example of how to implement app state restoration for the app that uses stateless widgets and gets its state from change notifiers, would be supremely helpful.

PR #568, opened 2020-11-12
Add support restoration api

nikitadol:
This is an example of what I meant in this issue https://github.com/rrousselGit/provider/issues/564

I accept any suggestions for improving the work
(I will write tests for this soon)

<details>
  <summary>Code sample</summary>
  
```dart
// ignore_for_file: public_member_api_docs, lines_longer_than_80_chars
import 'package:flutter/foundation.dart';
import 'package:provider/provider.dart';

/// This is a reimplementation of the default Flutter application using provider + [ChangeNotifier].

void main() {
  runApp(
    /// Providers are above [MyApp] instead of inside it, so that tests
    /// can use [MyApp] while mocking the providers
    RootRestorationScope(
      restorationId: 'root',
      child: MultiProvider(
        providers: [
          ChangeNotifierProvider(
            restorationId: 'conter',
            create: (_) => Counter(),
          ),
        ],
        child: MyApp(),
      ),
    ),
  );
}

/// Mix-in [DiagnosticableTreeMixin] to have access to [debugFillProperties] for the devtool
class Counter with ChangeNotifier, DiagnosticableTreeMixin, RestorationHandler {
  final RestorableInt _count = RestorableInt(0);

  int get count => _count.value;

  void increment() {
    _count.value++;
    notifyListeners();
  }

  /// Makes `Counter` readable inside the devtools by listing all of its properties
  @override
  void debugFillProperties(DiagnosticPropertiesBuilder properties) {
    super.debugFillProperties(properties);
    properties.add(IntProperty('count', count));
  }

  @override
  void restoreState(RestorationBucket oldBucket, bool initialRestore) {
    registerForRestoration(_count, 'count');
  }
}

  @override
    return const MaterialApp(
      home: MyHomePage(),
    );
  }
}

class MyHomePage extends StatelessWidget {
  const MyHomePage({Key key}) : super(key: key);

  @override
    return Scaffold(
      appBar: AppBar(
        title: const Text('Example'),
      ),
      body: Center(
        child: Column(
          mainAxisSize: MainAxisSize.min,
          mainAxisAlignment: MainAxisAlignment.center,
          children: <Widget>[
            const Text('You have pushed the button this many times:'),

            /// Extracted as a separate widget for performance optimization.
            /// As a separate widget, it will rebuild independently from [MyHomePage].
            ///
            /// This is totally optional (and rarely needed).
            /// Similarly, we could also use [Consumer] or [Selector].
            const Count(),
          ],
        ),
      ),
      floatingActionButton: FloatingActionButton(
        /// Calls `context.read` instead of `context.watch` so that it does not rebuild
        /// when [Counter] changes.
        onPressed: () => context.read<Counter>().increment(),
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ),
    );
  }
}

class Count extends StatelessWidget {
  const Count({Key key}) : super(key: key);

  @override
    return Text(

        /// Calls `context.watch` to make [Count] rebuild when [Counter] changes.
        '${context.watch<Counter>().count}',
        style: Theme.of(context).textTheme.headline4);
  }
}
```
  
</details>


rrousselGit:
I would suggest querying your DB instead. As the documentation of the state restoration explains, it isn't designed for "ambient state" – which is what provider is all about

nikitadol:
Some widgets can store their state, for example, [scroll position](https://github.com/flutter/flutter/blob/226532b1571c5d0fda22e6d050b85efa7bbbd069/packages/flutter/lib/src/widgets/scrollable.dart#L232-L246) or [even text values](https://github.com/flutter/flutter/blob/226532b1571c5d0fda22e6d050b85efa7bbbd069/packages/flutter/lib/src/widgets/editable_text.dart#L1236-L1253)

Why can't the provider do the same?

rrousselGit:
It's not that you _can't_ but that you _probably shouldn't_, as the documentation of the restoration API itself recommends.

The restoration API is made for UI state, which is composed of very light states like an ID or an offset.  
Providers are for complex/heavy business logic objects.

It makes more sense to use a DB instead.

nikitadol:
>very light states like an ID or an offset.

For example
User has a list (Todo)
And there is a search for this list, which is ralized through FIlterProvider
It just contains a field with the current query and/or filter

Why can't I save the state of this provider if it is needed to restore the UI?

rrousselGit:
Do you still want this?

If so, would you mind creating an issue? I'm not against having something similar. But we'd need to gather feedback from the community first, to understand what is desired here.

nikitadol:
Hi @rrousselGit, sorry for the long answer.

It would be very nice if there was a convenient API for this

Issue I already created https://github.com/rrousselGit/provider/issues/564

https://github.com/rrousselGit/provider/pull/568#issuecomment-743318694
I considered your option, but it is not very convenient and a lot of the same code

rrousselGit:
:wave: I'll close this. I don't plan on adding support for this for now, and it doesn't seem to get a lot of traction either

PR #574, opened 2020-11-23
feat!: null safety

### LINKED ISSUE #573: Migrate provider to null-safety
This is a tracker issue for migrating package:provider to null-safety.  More details coming soon.

felangel:
**WIP**
- opt into null safety (closes #573)
- still need to go back over a lot of the API surface
- mockito is still a dev_dependency so we will either need to wait for a null safety release of mockito or remove the dependency


mehmetf:
Thanks for this! I have briefly looked into it as well. I will leave some comments/questions for @rrousselGit in the code. There are certain variables that might need to be late initialized as opposed to nullable.

PS: Mockito migration will be painful. _any_ keyword is not null-safe compliant. I did not have the bandwidth to look at it deeply but two options are to use code generation to generate mock classes instead or explicitly override the methods with non-nullable params to make the parameters nullable.

lib/src/deferred_inherited_provider.dart
```diff
@@ -173,17 +173,17 @@ class _CreateDeferredInheritedProvider<T, R> extends _DeferredDelegate<T, R> {
...
 
 class _CreateDeferredInheritedProviderElement<T, R>
-    extends _DeferredDelegateState<T, R,
-        _CreateDeferredInheritedProvider<T, R>> {
+    extends _DeferredDelegateState<T?, R,
+        _CreateDeferredInheritedProvider<T?, R>?> {
   bool _didBuild = false;
 
-  T _controller;
+  T? _controller;
```
mehmetf, line 180: Is _controller nullable or just late initialized? @rrousselGit

lib/src/inherited_provider.dart
```diff
@@ -550,17 +550,17 @@ typedef _SelectorAspect<T> = bool Function(T value);
...
 
-abstract class _DelegateState<T, D extends _Delegate<T>> {
-  _InheritedProviderScopeElement<T> element;
+abstract class _DelegateState<T, D extends _Delegate<T?>?> {
+  late _InheritedProviderScopeElement<T> element;
 
   T get value;
 
-  D get delegate => element.widget.owner._delegate as D;
+  D? get delegate => element.widget.owner._delegate as D?;
```
mehmetf, line 563: I believe delegate is not supposed to be nullable.

lib/src/inherited_provider.dart
```diff
@@ -602,23 +602,23 @@ class _CreateInheritedProvider<T> extends _Delegate<T> {
...
 
 class _CreateInheritedProviderState<T>
-    extends _DelegateState<T, _CreateInheritedProvider<T>> {
-  VoidCallback _removeListener;
+    extends _DelegateState<T?, _CreateInheritedProvider<T?>?> {
+  VoidCallback? _removeListener;
   bool _didInitValue = false;
-  T _value;
-  _CreateInheritedProvider<T> _previousWidget;
+  T? _value;
```
mehmetf, line 615: `_didInitValue` made me think _value can be late initialized. I ran into problems primarily due to _create function being optional. _update takes in _value which then causes a read before a write.

Is null an acceptable `_value`? @rrousselGit

felangel:
> Thanks for this! I have briefly looked into it as well. I will leave some comments/questions for @rrousselGit in the code. There are certain variables that might need to be late initialized as opposed to nullable.
> 
> PS: Mockito migration will be painful. _any_ keyword is not null-safe compliant. I did not have the bandwidth to look at it deeply but two options are to use code generation to generate mock classes instead or explicitly override the methods with non-nullable params to make the parameters nullable.

No problem and yes, it's still a work in progress. I will continue to refactor and would appreciate any/all feedback 👍

rrousselGit:
I don't think it makes sense to allow `StreamProvider<int>.value(value: null)`, nor does it make sense to _not_ have an initial value

The initialValue likely has to be required now. The current behavior is to return `null` when streams are loading, which is not feasible if the stream content is non-nullable.

rrousselGit:
I started going through the PR, but there are too many changes I want to make, so I'll summarize it here:

It is critical that instead of:

```dart
T? watch<T>()
```

for `context.watch`, we have:

```dart
T watch<T>()
```

(same thing what `Provider.of` and others)

This causes a cascade of changes necessary to make it work, including my previous comment about making `StreamProvider.initialData` required

rrousselGit:
For the mockito migration, I didn't realize how painful it would be.

In that case, it may make sense to stop using mockito. I'm not fond of the idea of using code-generation to write tests.

mehmetf:
+1 to making breaking changes so the library is not just migrated but also carries the right null-safety semantics.

The downside is the timelines. We want to migrate provider within the next month or so since it is being used by _many_ applications in Google and we cannot start migrating the applications before the dependencies are migrated. Based on this, I will fork it and do a "surface migration" (essentially this PR). Then, we can migrate to the proper one when Remi has time to push it through. That should take off the timeline pressure.

@felangel, Remi has the final say on this but I agree with the direction you are taking. If you can get the tests working (with or without mockito) in a null-safe environment, that would still be a huge help to Remi. Then this PR can be submitted and we can start making additional null-safety tweaks in smaller PRs.

felangel:
I rebased and resolved merge conflicts and updated the controller to be late initialized. Going to pause now but @rrousselGit feel free to push additional fixes/updates.

> I started going through the PR, but there are too many changes I want to make, so I'll summarize it here:
> 
> It is critical that instead of:
> 
> ```dart
> T? watch<T>()
> ```
> 
> for `context.watch`, we have:
> 
> ```dart
> T watch<T>()
> ```
> 
> (same thing what `Provider.of` and others)
> 
> This causes a cascade of changes necessary to make it work, including my previous comment about making `StreamProvider.initialData` required

I will address this later today but it would be super helpful if someone can help update tests especially if we're going to remove `mockito`.

mehmetf:
I would rather leave the modification of `T? watch<T>()` to `T watch<T>()` to Remi since that kind of change requires a redesign of the APIs in the direction that Remi wants to go. 

As such, I think the lib/* portion of this PR can be submitted as is. What we need to do is fix the tests somehow.

felangel:
> I would rather leave the modification of `T? watch<T>()` to `T watch<T>()` to Remi since that kind of change requires a redesign of the APIs in the direction that Remi wants to go.
> 
> As such, I think the lib/* portion of this PR can be submitted as is. What we need to do is fix the tests somehow.

I’m fine either way. @rrousselGit let us know how you prefer to proceed 👍

mehmetf:
We might not need to migrate off of Mockito. The codegen solution isn't so bad and it will be the canonical way to create mocks in a null-safe world. 

https://github.com/dart-lang/mockito/blob/master/NULL_SAFETY_README.md

Wikiwix:
> I don't think it makes sense to allow `StreamProvider<int>.value(value: null)`, nor does it make sense to _not_ have an initial value
> 
> The initialValue likely has to be required now. The current behavior is to return `null` when streams are loading, which is not feasible if the stream content is non-nullable.

In my opinion there should be some mechanism to allow differentiating uninitialised values from a default value (if there is one that makes sense.\
Having to reserve a specific value as the *uninitialised* value seems like a bad idea for me.

[In an old PR](https://github.com/rrousselGit/provider/pull/126#issuecomment-504495642) someone asked for the usecase to differentiate and your answer was to use `null` for that. (Unfortunately I cannot find that issue anymore :( )

Don't get me wrong, I think your "*nnbd approach*" for `StreamProvider seems the best default, but perhaps add some other kind of `Consumer` that knows about the `Stream`'s state to allow a clean solution for the specific usecase.

rrousselGit:
> Having to reserve a specific value as the uninitialised value seems like a bad idea for me.

The recommendation would be to expose an AsyncSnashot<T> instead of just <T>

Wikiwix:
> > Having to reserve a specific value as the uninitialised value seems like a bad idea for me.
> 
> The recommendation would be to expose an AsyncSnashot instead of just

Oh that sounds good. So all I have to say for the discussion is that I second your opinion on going with `T watch<T>()`.

Regarding your solution: I did search for it and I thought about ways to build it (even looked at the implementation of `StreamBuilder`), but I am at a loss on how to make use of an `AsyncSnapshot` with `Provider` to expose a `Stream`.
Could you perhaps point me in the right direction?

And I guess this should be added to the Documentation of `StreamProvider` at some point 🙈 . If I understood how, I might even provide a PR :).

vishna:
So I was wondering, now with null safety, if we have the following:

```dart
final myValue = Provider.of<T>(context); // should throw exception if value is not there
```

will this not throw anymore?

```dart
final myMaybeValue = Provider.of<T?>(context); // doesn't throw, return null if value is not there
```

rrousselGit:
This will still throw

But I guess we could modify the behavior so that:
- if the generic type is nullable, do not throw
- if it isn't, throw

rrousselGit:
I have locally migrated all the files, including mockito (using the manual approach), but am currently stuck on one issue

I get a compilation error with no explanation:

```dart
remirousselet@TABLT0327-RemiRousselet provider % flutter analyze .
Analyzing provider...                                                   

   info • The library 'package:mockito/mockito.dart' is legacy, and should not be imported into a
          null safe library • test/common.dart:6:8 • import_of_legacy_library_into_null_safe

1 issue found. (ran in 2.0s)
remirousselet@TABLT0327-RemiRousselet provider % flutter test     
00:01 +0: ... /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart                 Error: Cannot run with sound null safety, because the following dependencies
don't support null safety:

 - package:mockito

For solutions, see https://dart.dev/go/unsound-null-safety
00:01 +0 -1: loading /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart [E]      
  Failed to load "/Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart":
  Compilation failed
  Test: /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart
  Shell: /Users/remirousselet/dev/flutter/bin/cache/artifacts/engine/darwin-x64/flutter_tester
  
  
00:01 +0 -1: Some tests failed.  
```


As you can see, there's nothing really helpful here

This is especially surprising because I can start the example folder

@mehmetf Do you know what could be the problem or how to debug this issue?

Wikiwix:
> I have locally migrated all the files, including mockito (using the manual approach), but am currently stuck on one issue
> 
> I get a compilation error with no explanation:
> 
> ```
> remirousselet@TABLT0327-RemiRousselet provider % flutter analyze .
> Analyzing provider...                                                   
> 
>    info • The library 'package:mockito/mockito.dart' is legacy, and should not be imported into a
>           null safe library • test/common.dart:6:8 • import_of_legacy_library_into_null_safe
> 
> 1 issue found. (ran in 2.0s)
> remirousselet@TABLT0327-RemiRousselet provider % flutter test     
> 00:01 +0: ... /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart                 Error: Cannot run with sound null safety, because the following dependencies
> don't support null safety:
> 
>  - package:mockito
> 
> For solutions, see https://dart.dev/go/unsound-null-safety
> 00:01 +0 -1: loading /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart [E]      
>   Failed to load "/Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart":
>   Compilation failed
>   Test: /Users/remirousselet/dev/rrrouselgit/provider/test/consumer_test.dart
>   Shell: /Users/remirousselet/dev/flutter/bin/cache/artifacts/engine/darwin-x64/flutter_tester
>   
>   
> 00:01 +0 -1: Some tests failed.  
> ```
> 
> As you can see, there's nothing really helpful here
> 
> This is especially surprising because I can start the example folder
> 
> @mehmetf Do you know what could be the problem or how to debug this issue?

Your *manual approach* does not seem to be recognised by the analyzer.

Does `dart --no-sound-null-safety test` work?
Because for mixed mode you explicitly need to tell all the tools that. I assume that the example does not use `mockito` and therefore works?

rrousselGit:
> Does dart --no-sound-null-safety test work?

How does that work with `flutter test`? The flag isn't recognized/documented

Wikiwix:
> > Does dart --no-sound-null-safety test work?
> 
> ~How does that work with `flutter test`? The flag isn't recognized/documented~
> 
> Apparently putting the flag _after_ "test" works, even if undocumented
> So `flutter test --no-sound-null-safety`
> 
> Which works, thanks!

Great to hear, just checked again (whether I mixed the order up), but for `dart` the flag **has** to be *before*…

Wikiwix:
@rrousselGit Some other "fix" I just stumbled upon: https://dart.dev/null-safety/unsound-null-safety#testing-or-running-mixed-version-programs

Probably the best approach still is to "properly" migrate `mockito` I guess.

mehmetf:
FYI @srawlins Do you know why tests depending on mockito need to run in mixed mode? I believe mockito package is already migrated.

srawlins:
> FYI @srawlins Do you know why tests depending on mockito need to run in mixed mode? I believe mockito package is already migrated.

There is not yet a published null safe release of mockito. I'm continuing to work on it this week.

lib/src/provider.dart
```diff
@@ -287,42 +286,22 @@ The context used was: $context
...
+            _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>?;
         return false;
       });
     } else {
       inheritedElement = context.getElementForInheritedWidgetOfExactType<
-          _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>;
+          _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>?;
     }
 
     if (inheritedElement == null) {
```
rokob, line 307: Since the return type here is T? can you just return inheritedElement rather than throwing this exception? Otherwise the return type should logically be T as this will never return null due to the throw.

rrousselGit:
I've finished the migration 🎉 

Feel free to review it. If not, I will merge it and publish it tomorrow

mehmetf:
This is great! I am going to patch this in and test it in Google to give you a better signal for correctness.

mehmetf:
aaand I immediately ran into a problem. `ValueListenableProvider` is quite popular with some teams. There are ~500 references to the `value` constructor, most of it from tests.

I also see some patterns like this where only value parameter is supplied but no `child`:

```dart
return MultiProvider(
      providers: [
        ValueListenableProvider.value(value: foo),
        ValueListenableProvider.value(value: bar),
        Provider.value(value: baz),
...
```

How do we migrate such cases?

mehmetf:
I am also curious why this was deprecated. Is that API cleanup? It seems like a useful feature for teams and it doesn't look like a bad API. I would think someone will likely create a helper class that does exactly the same thing (and possibly incorrectly). @rrousselGit

rrousselGit:
Interesting. Do you mind sharing what is the use-case?

This provider was my mistake. I added it without any real use-case in mind, in a bundle with ChangeNotifierProvider.
Its default constructor is basically unusable and you are the first person telling me that this provider is used.

In the end, only the `.value` is potentially useful, which is very easily to implement on your own:

```dart
import 'package:nested/nested.dart';

class Example<T> extends SingleChildStatelessWidget {
  Example({
    Key? key,
    required this.valueListenable
    Widget? child,
  }): super(key: key, child: child);

  final ValueListenable<T> valueListenable;

  @override
  Widget buildWithChild(BuildContext context, Widget? child) {
    return ValueListenableBuilder<T>(
      valueListenable: valueListenable,
      builder: (context, value, _) => Provider<T>.value(value: value, child: child),  
    );
  }
}
```

mehmetf:
I can definitely find out about the use case. It seems like it is team based. ~460 out of those 500 references is coming from one team. 

Even the default constructor is being used; for instance in here:

```dart
      final key = GlobalKey<MyWidgetState>();
      final valueNotifier = ValueNotifier(_ProvidedObject('hello'));
      await tester.pumpWidget(
        MultiProvider(
          providers: [
            ValueListenableProvider(create: (_) => valueNotifier),
          ],
          child: MyWidget(
            key: key,
            ...
          ),
        ),
      );
      key.currentState.resume();
      valueNotifier.value = _ProvidedObject('world');
      await tester.pumpAndSettle();
      expect(fakeTrigger.wasDidChangeDependenciesCalled, isTrue);
      expect(fakeTrigger.capturedProvidedObject.value, 'world');
```

I am fairly sure all this can be rewritten in a way that does not use this provider. However, it would make the migration quite painful since it is a code structure change and each case needs to be reviewed and rewritten. For instance, if I understand correctly, the use case of:

```dart
return MultiProvider(
      providers: [
        ValueListenableProvider.value(value: foo),
        ValueListenableProvider.value(value: bar),
        Provider.value(value: baz),
...
```

would have to be changed to a normal provider. Then, we need to find out where `foo` and `bar` are being used in the application code and make structural changes there by using `ValueListenableBuilder` to make sure the same semantics apply. Is that correct? 

If so, it would take weeks. null-safety is already a risky migration with critical timelines so I would much rather avoid tacking this on. What do you think?

rrousselGit:
> ```dart
>       final valueNotifier = ValueNotifier(_ProvidedObject('hello'));
>       await tester.pumpWidget(
>         MultiProvider(
>           providers: [
>             ValueListenableProvider(create: (_) => valueNotifier),
> ```

You could safely use `.value` in this scenario

> . Then, we need to find out where foo and bar are being used in the application code and make structural changes there by using ValueListenableBuilder to make sure the same semantics apply. Is that correct?

No need. You can use the snippet I gave above instead. So you'd have:

```dart
return MultiProvider(
      providers: [
        Example(value: foo),
        Example(value: bar),
        Provider.value(value: baz),
```

rrousselGit:
Could you search for usages of the default constructor that cannot be replaced by the `.value` constructor?

If all the usages of `ValueListenableProvider` are using `.value`, I don't mind deprecating _only_ the default constructor and leaving the `.value`

mehmetf:
Thanks! I have a PR to change the default ctr usage to `.value`. The tests pass so we can safely deprecate it.

mehmetf:
If you resurrect `.value` constructor, I can test the null-safety migration PR as well and give you feedback.

mehmetf:
I put the ValueListenableProvider back in, removed the default ctr and tested in Google. The only failures were the known ones due to initialData now being required. Since most of those are building complex objects asynchronously, the proper solution is to pass null explicitly. The tests all passed so the migration looks good!

rrousselGit:
I've published a new nnbd version that includes `ValueListenableProvider.value`

rrousselGit:
>  Since most of those are building complex objects asynchronously, the proper solution is to pass null explicitly. 

Not necessary as this is a bigger change, but my recommendation would be to use `AsyncSnapshot` (or a similar object):

```dart
StreamProvider<AsyncSnapshot<int>>(
  initialData: const AsyncSnapshot.waiting(),
  create: (context) async* {},
  catchError: (context, error) => AsyncSnapshot.withError(ConnectionState.active, error!),
)
```

mehmetf:
> My recommendation would be to use AsyncSnapshot

Ultimately it boils down to what pattern to use to represent initial state of an object that requires async initialization:

`Foo` type with a special case for not available.

or

`Foo?` type with `null` to represent not available.

I am not sure if there's a consensus on this yet. The code still needs to systemically check for the not available state and promote the object one way or another.

mehmetf:
@rrousselGit The `ValueListenableProvider` change is breaking in a subtle way. While the change preserves the current behavior _eventually_ , the timing changes. You are currently listening to changes in `ValueListenable` and [setting the state synchronously](https://github.com/rrousselGit/provider/blob/master/lib/src/value_listenable_provider.dart#L80-L81) (then scheduling a frame). With the change to `ValueListenableBuilder`, the changed value is propagated in the next frame.

We had a client test break because they had another listener attached to `ValueListenable`. This listener triggers a microtask which reads the value of the listenable via its provider. In the old method, by the time the microtask executed, the value was already updated in the provider. Now, there's no such guarantee. I believe what they are doing is wrong. Subscribing to a value listenable change stream and then using a different mechanism to read its value is asking for trouble so I advised them to change this.

I wanted to let you know so you can document this subtle change.

rrousselGit:
You mean that with such code:

```dart
final notifier = ValueNotifier(0);

ValueListenableProvider.value(
  value: notifier,
  builder: (context, child) {
    return RaisedButton(
      onPressed: () {
        print(context.read<int>());
        notifier.value++;
        print(context.read<int>());
      },
    );
  },
)
```

before, it used to print:

```
0
1
1
```

and now it prints:

```
0
1
0
```

?

mehmetf:
Yes, in a way.

The code you gave used to print `0  1`. Now it will print `0  0`. I am not sure where the third print is coming from since all the prints are in the gesture handler.

rrousselGit:
Sorry ignore the 2nd print, I forgot to add it in the example

You're correct. Thanks!

lib/src/inherited_provider.dart
```diff
@@ -667,37 +667,37 @@ class _CreateInheritedProviderState<T>
...
 
-    element._isNotifyDependentsEnabled = false;
-    _removeListener ??= delegate.startListening?.call(element, _value);
-    element._isNotifyDependentsEnabled = true;
+    element!._isNotifyDependentsEnabled = false;
+    _removeListener ??= delegate.startListening?.call(element!, _value as T);
+    element!._isNotifyDependentsEnabled = true;
     assert(delegate.startListening == null || _removeListener != null);
-    return _value;
+    return _value as T;
```
felangel, line 692: I think this is incorrect. Previously if a value was null we would throw a `ProviderNotFoundException` however, with this change I believe we will instead have a `_CastError` type 'Null' is not a subtype of type 'T' in type cast.
```suggestion
    return _value as T?;
```

rrousselGit: We cannot return null. That would make context.watch return T?

felangel: Thoughts on something like:

```dart
if (_value == null) throw ProviderNotFound(...);
return _value as T;
```

I don't think the `CastError` should be propagated to developers. It should be handled internally and surfaced as the previous `ProviderNotFound` exception. Thoughts?

rrousselGit: Null may be a valid value

To begin with, "create" should not throw.
It may need a custom error ( i don't believe ProviderNotFound fits here), but it definitely should throw

felangel: Yeah good point.

As I mentioned above, I think it's valid to perform a read from within a create which could potentially fail:

```dart
Provider<String>(
  create: (context) {
    return '${context.read<int>()}';
  },
  child: ...
);
```

If the upstream provider is not found, `create` will throw and I'm fine with a custom error (maybe `ProviderCreateException`?).

The main things imo are:

- The case described above should be able to be handled by other providers
- The error shown to the developer should be easy to understand and point to the root cause

felangel: The changes in https://github.com/rrousselGit/provider/pull/574/commits/4838869adc2be94e02c903b4057b76d067bad61f address this issue and make it backward compatible with the previous behavior. @rrousselGit can you please take a look and let me know what you think? I verified that the flutter_bloc test suite now passes, thanks! 🙏

lib/src/inherited_provider.dart
```diff
@@ -561,16 +561,16 @@ abstract class _Delegate<T> {
 }
 
 abstract class _DelegateState<T, D extends _Delegate<T>> {
-  _InheritedProviderScopeElement<T> element;
+  _InheritedProviderScopeElement<T>? element;
 
-  T get value;
+  T? get value;
```
rrousselGit, line 566: It does not make sense for the interface to make the getter nullable

lib/src/inherited_provider.dart
```diff
@@ -667,37 +667,41 @@ class _CreateInheritedProviderState<T>
...
-                _debugPreviousIsInInheritedProviderUpdate;
+                _debugPreviousIsInInheritedProviderUpdate!;
             return true;
           }());
         }
 
         assert(() {
-          delegate.debugCheckInvalidValueType?.call(_value);
+          if (_value is T) {
+            delegate.debugCheckInvalidValueType?.call(_value as T);
```
rrousselGit, line 683: I would prefer starting this getter with:

```dart
if (_value is! T) {
  throw StateError(
     'Tried to read a provider that threw during the creation of its value.\n'
      'Fix the exception in the provider to this error',
  );
}
```

test/inherited_provider_test.dart
```diff
@@ -2408,15 +2400,47 @@ DeferredInheritedProvider<int, int>(controller: 42, value: 24)'''),
...
+    final dynamic exception = tester.takeException();
+    expect(
+      exception,
+      isA<ProviderNotFoundException>().having(
+        (e) => e.valueType,
+        'valueType',
+        String,
+      ),
+    );
+  });
```
rrousselGit, line 2423:
```suggestion
  testWidgets('throws ProviderNotFoundException if lookup fails within create',
      (tester) async {
    await tester.pumpWidget(
      Provider(
        lazy: false,
        create: (context) {
          context.read<String>();
          return 42;
        },
        child: const SizedBox(),
      ),
    );

    final dynamic exception = tester.takeException();

    expect(
      exception,
      isA<ProviderNotFoundException>().having(
        (e) => e.valueType,
        'valueType',
        String,
      ),
    );
  });
```

test/inherited_provider_test.dart
```diff
@@ -2408,15 +2400,47 @@ DeferredInheritedProvider<int, int>(controller: 42, value: 24)'''),
     expect(find.text('true true'), findsOneWidget);
     expect(buildCount, 4);
   });
+
+  testWidgets('throws ProviderNotFoundException if lookup fails within create',
```
rrousselGit, line 2404: We need a test for what happens when we call `context.watch` after a provider threw

test/inherited_provider_test.dart
```diff
@@ -2408,15 +2400,47 @@ DeferredInheritedProvider<int, int>(controller: 42, value: 24)'''),
...
+    );
+  });
+
+  testWidgets('propagates exceptions within create', (tester) async {
+    final e = Exception('oops');
+    await tester.pumpWidget(Provider(
+      lazy: false,
+      create: (context) => throw e,
+      child: const SizedBox(),
+    ));
```
rrousselGit, line 2431:
```suggestion

    await tester.pumpWidget(
      Provider(
        lazy: false,
        create: (context) => throw e,
        child: const SizedBox(),
      ),
    );

```

mehmetf:
@rrousselGit Another breaking change discovered for `ValueListenableProvider`. The assert gets triggered for `Provider` that disallows a value to be a `Listenable` or `Stream`. Previously, it was legal to construct a `ValueListenableProvider<T>` where T itself was a `Listenable`. Now, that's not allowed.

To be fair, I think it is a code smell to wrap a Listenable in another Listenable. The team that was doing that had a service that published updates. The service itself could change due to a different signal in the app which is why they wrapped it in a `ValueListenable`. I advised the team to encapsulate that signal explicitly and make the service part of that encapsulation. However, this should be documented as it is a breaking change.

felangel:
@rrousselGit can you please review the changes and let me know if there's anything else you'd like me to adjust? I would love to have an updated null safety release which includes these changes to unblock the migration of the flutter_bloc package. Thanks 🙏

rrousselGit:
Looking good, thanks!
I'll publish a new version

rrousselGit:
Alright time to merge this!

I won't immediately release a 5.0.0 (without the --nullsafety) because there's a WIP feature that should be released before march that needs to be bundled with this version.

Issue #576, opened 2020-11-25
WatchContext extension cannot be used in didChangeDependencies with AssertionException


PlugFox:
From comment:
> Calling this method is equivalent to calling:
> ```dart
> Provider.of<T>(context)
> ```

WatchContext extension is shortcut for Provider.of(context), but nope, WatchContext extension cannot be used in didChangeDependencies, something like this:
```dart
class _ScreenState extends State<Screen> {
  int _currentData;

  @override
  void didChangeDependencies() {
    super.didChangeDependencies();
    _currentData = context.watch<int>();
  }

  @override
  Widget build(BuildContext context) {
    //final _currentData = context.watch<int>();
    return Text((_currentData ?? 0).toString());
  }
}
```

This lines
https://github.com/rrousselGit/provider/blob/9c4087abaa753ba5bc15492cf639d97720048cfb/lib/src/provider.dart#L639-L651
must be like this:
```dart
assert(
            owner.debugBuilding || // <= ADDED
            widget is LayoutBuilder ||
            widget is SliverWithKeepAliveWidget ||
            debugDoingBuild ||
            debugIsInInheritedProviderUpdate,
        '...');
```

Exception listing:
```
════════ Exception caught by widgets library ═══════════════════════════════════════════════════════
The following assertion was thrown building _InheritedProviderScope<int>:
Tried to use `context.watch<int>` outside of the `build` method or `update` callback of a provider.

This is likely a mistake, as it doesn't make sense to rebuild a widget when the value
obtained changes, if that value is not used to build other widgets.

Consider using `context.read<int> instead.
'package:provider/src/provider.dart':
Failed assertion: line 620 pos 9: 'widget is LayoutBuilder ||
            widget is SliverWithKeepAliveWidget ||
            debugDoingBuild ||
            debugIsInInheritedProviderUpdate'

The relevant error-causing widget was: 
  Provider<int> file:///C:/git/flutter/provider_watch_feature/lib/main.dart:27:49
When the exception was thrown, this was the stack: 
#2      WatchContext.watch (package:provider/src/provider.dart:620:9)
#3      _ScreenState.didChangeDependencies (package:provider_watch_feature/main.dart:69:28)
#4      StatefulElement._firstBuild (package:flutter/src/widgets/framework.dart:4833:11)
#5      ComponentElement.mount (package:flutter/src/widgets/framework.dart:4649:5)
#6      Element.inflateWidget (package:flutter/src/widgets/framework.dart:3615:14)
```

[...]

PR #577, opened 2020-11-25
Using watch extension with didChangeDependencies

PlugFox:
https://github.com/rrousselGit/provider/issues/576


rrousselGit:
Closing this for now.

That does not mean nothing will change here. I will very likely remove the assert in the near future.
But the change proposed here is not what is desired (as shown by the failing CI)

You're free to either fix the CI (by removing the assert completely and fixing the tests), or wait until some time until it's higher on my todo-list

PR #595, opened 2021-01-08
Likely adding new provider needs hot restart

salvaterra:
This is the most likely scenario and trigger this error

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -421,9 +421,11 @@ class ProviderNotFoundException implements Exception {
     return '''
 Error: Could not find the correct Provider<$valueType> above this $widgetType Widget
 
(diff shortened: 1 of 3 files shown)


lib/src/provider.dart
```diff
@@ -421,7 +421,8 @@ class ProviderNotFoundException implements Exception {
     return '''
 Error: Could not find the correct Provider<$valueType> above this $widgetType Widget
 
-This likely happens because you used a `BuildContext` that does not include the provider
+This likely happens because you created a new provider but did not performed a 'Hot Restart'
```
rrousselGit, line 424: I'd instead include that in the "common scenarios" list, making sure to explain that this is for when adding providers inside the `main` function

PR #596, opened 2021-01-08
Remove hasListeners assertion from ListenableProvider

### LINKED ISSUE #592: Incorrect assertion failure about an unused Listenable?
**Describe the bug**
This code:

```dart
ChangeNotifierProvider(
  create: (context) {
    final notifier = ValueNotifier<String>(null);
    notifier.addListener(() {
      final value = notifier.value;
      // ... do stuff based on value
    });
    return notifier;
  },
child: someChild,
};
```

runs into this assertion: https://github.com/rrousselGit/provider/blob/v4.3.2+2/lib/src/listenable_provider.dart#L43

I'm not sure whether the fact that a `Listenable` has listeners is a good way of asserting it is a 'fresh' instance, like in my example above - it is clearly a new one that is managed by `provider`, but I just want to attach a listener to it right away.

There are quite a few ways I can do it in anyway. The one that is suggested in the error message (using `Provider.value` instead) is very verbose, requires the developer to explicitly create and then dispose of the instance, which may require implementing additional code, like using a `StatefulWidget` instead of `StatelessWidget` and using `initState` and `dispose` to manage the state themselves - which is a lot of (in my opinion) unnecessary code, forced upon the developer because of a single assertion that really isn't checking what is should (as mentioned, the fact that there are listeners doesn't mean it is not a fresh instance).

~~Another way, the one I used eventually, is using an intermediate `Builder` widget:~~

```dart
ChangeNotifierProvider(
  create: (context) => ValueNotifier<String>(null),
  child: Builder(
    builder: (context) {
      final notifier = Provider.of<ValueNotifier<String>>(context);
      notifier.addListener(() {
        final value = notifier.value;
        // ... do stuff based on value
      });
      return someChild;
    },
  ),
);
```
but it still is strange code that requires me to look up a notifier that I have just created.

**Edit**: this solution is actually broken as it will result in a new listener being added on each rebuild.

Or is it bad what I'm doing in the first place (wanting to create an instance with a listener)?

**Expected behavior**

I can create a new `ListenableProvider` (or subclasses) with an attached listener and it still works, without ceremony ;d

**Side note**

First let me thank you for `provider`, it has been great for my team and a huge relief to not have to deal with `InheritedWidget` verbosity all the time myself. However...

Lately I've been getting the impression that `provider` is getting more and more opinionated with each release, and, I don't know, harder to use? Assertions like the one mentioned in this issue, or the ones checked in `context.read` and `context.watch` are a bit too restrictive and are not right some of the time. I know I can still use `Provider.of` in such cases, but it makes the API surface bigger and more confusing, and my juniors at work have trouble to get when to use what ...

Then there is the fact that I can't provide a `Stream` using simply `Provider`, I need to use `StreamProvider` and then my widget rebuilds based on stream changes when I used e.g. `context.watch` - but I don't want that all the time, e.g. I have my own BLoC pattern implementation where the `Bloc` class is a `Stream` but I don't want it to be treated as a `Stream` by `provider`, I know better in this particular case. (I know I can disable the check, but it is done with a static field and is an 'all-or-nothing' solution - it disables checks for the whole app at once.)

Is it really necessary to have all those checks/assertions done by the framework, and have it throw (in debug mode)? It can't check everything for my anyway, and in my personal experience with our app the assertions are wrong most of the time, and simply rather not that helpful, not to say ... frustrating?

wujek-srujek:
This fixes #592 and allows the following to be valid code:
```dart
ChangeNotifierProvider(
  create: (context) {
    final notifier = ValueNotifier<String>('initialValue');
    notifier.addListener(() {
      final value = notifier.value;
      // ... do stuff based on value
    });
    return notifier;
  },
  child: Text('Abc'),
};
```

### DIFF
--- a/test/listenable_provider_test.dart
+++ b/test/listenable_provider_test.dart
@@ -27,18 +27,22 @@ void main() {
           listenable);
     });
     testWidgets(
-      'asserts that the created notifier has no listener',
+      'asserts that the created notifier can have listeners',
       (tester) async {
+        final key = GlobalKey();
         final notifier = ValueNotifier(0)..addListener(() {});
 
         await tester.pumpWidget(
           ListenableProvider(
             create: (_) => notifier,
-            child: const TextOf<ValueNotifier<int>>(),
+            child: Container(key: key),
           ),
         );
 
-        expect(tester.takeException(), isAssertionError);
+        expect(
+          Provider.of<ValueNotifier<int>>(key.currentContext, listen: false),
+          notifier,
+        );
       },
     );
 
--- a/lib/src/listenable_provider.dart
+++ b/lib/src/listenable_provider.dart
@@ -35,35 +35,6 @@ class ListenableProvider<T extends Listenable> extends InheritedProvider<T> {
           startListening: _startListening,
           create: create,
           dispose: dispose,
-          debugCheckInvalidValueType: kReleaseMode
-              ? null
-              : (value) {
-                  if (value is ChangeNotifier) {
-                    // ignore: invalid_use_of_protected_member
-                    assert(!value.hasListeners, '''
-The default constructor of ListenableProvider/ChangeNotifierProvider
-must create a new, unused Listenable.
-
-If you want to reuse an existing Listenable, use the second constructor:
-
-- DO use ChangeNotifierProvider.value to provider an existing ChangeNotifier:
-
-MyChangeNotifier variable;
-ChangeNotifierProvider.value(
-  value: variable,
-  child: ...
-)
-
-- DON'T reuse an existing ChangeNotifier using the default constructor.
-
-MyChangeNotifier variable;
-ChangeNotifierProvider(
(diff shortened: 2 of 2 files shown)


test/listenable_provider_test.dart
```diff
@@ -26,21 +26,6 @@ void main() {
       expect(Provider.of<ChangeNotifier>(key.currentContext, listen: false),
           listenable);
     });
-    testWidgets(
-      'asserts that the created notifier has no listener',
```
wujek-srujek, line 30: Is it enough to remove the test for the assertion (which I think is enough), or should I create a dedicated test which tests that a `ListenableProvider` with a listener doesn't fail?

rrousselGit: Make sure to add a test for this specific use-case. Thanks!

wujek-srujek: Done.

PR #617, opened 2021-03-06
Change "over the time" to "over time"

arthurwolf:
It's more correct English (I *think*) / less odd. Sorry if I'm wrong about this.


rrousselGit:
Oh good catch, thanks!

PR #626, opened 2021-03-22
ChangeNotifier isn't O(N2) anymore, as can be seen on the docs.

feinstein:


README.md
```diff
@@ -456,10 +456,10 @@ initState() {
 }
 ```
 
-This is not allowed, because the modification is immediate.
+This is not allowed, because the modification has to be immediate.
```
rrousselGit, line 459:
```suggestion
This is not allowed, because the state update is synchronous.
```

rrousselGit:
Thanks!

PR #648, opened 2021-06-11
add plugin and usage a skills

CNAD666:
**add provider plugin**
  - one-click generation of necessary folders, files, template codes, etc. for each page
  - support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
  - Support input **provider** prompt to generate quick code snippets

**add provider usage a skills**

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,7 @@ See also:
 
 **In order to further improve your production efficiency, we have also prepared plug-ins for you**
 
-- provider_template：[Android Studio/Intellij插件](https://plugins.jetbrains.com/plugin/16821-flutter-provider)
+- provider_template：[Android Studio/Intellij   Plugin](https://plugins.jetbrains.com/plugin/16821-flutter-provider)
   - one-click generation of necessary folders, files, template codes, etc. for each page
   - support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
   - Support input **provider** prompt to generate quick code snippets
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ See also:
 
 - provider_template：[Android Studio/Intellij   Plugin](https://plugins.jetbrains.com/plugin/16821-flutter-provider)
   - one-click generation of necessary folders, files, template codes, etc. for each page
-  - support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
+  - support Wrap Widget: Consumer, Selector, ChangeNotifierProvider
   - Support input **provider** prompt to generate quick code snippets
 
 ### Some usage
@@ -97,22 +97,24 @@ class CounterPage extends StatelessWidget {
   Widget build(BuildContext context) {
     return ChangeNotifierProvider(
       create: (BuildContext context) => CounterProvider(),
-      child: Scaffold(
-        appBar: AppBar(title: Text('Provider')),
-        body: Center(
-          child: Consumer<ProEasyCounterProvider>(
-            builder: (context, provider, child) {
-              return Text('count ${provider.count} ');
-            },
+      builder: (BuildContext context, Widget? child) {
+        return Scaffold(
+          appBar: AppBar(title: Text('Provider')),
+          body: Center(
+            child: Consumer<CounterProvider>(
+              builder: (context, provider, child) {
+                return Text('count ${provider.count} ');
+              },
+            ),
           ),
-        ),
-        floatingActionButton: FloatingActionButton(
-          onPressed: () {
-            Provider.of<ProEasyCounterProvider>(context, listen: false).increment();
-          },
-          child: Icon(Icons.add),
-        ),
-      ),
+          floatingActionButton: FloatingActionButton(
+            onPressed: () {
+              Provider.of<CounterProvider>(context, listen: false).increment();
+            },
+            child: Icon(Icons.add),
+          ),
+        );
+      },
     );
   }
 }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -78,6 +78,95 @@ See also:
...
+`Provider.of` is a little troublesome to use, we can use some tips to optimize it, ChangeNotifierProvider will store the XxxProvider object, here, we can also store it for subsequent use
+
+```dart
+class CounterPage extends StatelessWidget {
+  final provider = CounterProvider();
+
+  @override
+  Widget build(BuildContext context) {
+    return ChangeNotifierProvider(
+      create: (BuildContext context) => provider,
```
rrousselGit, line 141: This is anti-pattern.

The create function should not reuse an existing variable and instead create a new state.

README.md
```diff
@@ -78,6 +78,95 @@ See also:
...
+### Plugin
+
+**In order to further improve your production efficiency, we have also prepared plug-ins for you**
+
+- provider_template：[Android Studio/Intellij   Plugin](https://plugins.jetbrains.com/plugin/16821-flutter-provider)
+  - one-click generation of necessary folders, files, template codes, etc. for each page
+  - support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
+  - Support input **provider** prompt to generate quick code snippets
+
+### Some usage
```
rrousselGit, line 90: I am not sure what this part adds over the existing documentation

README.md
```diff
@@ -78,6 +78,95 @@ See also:
 
 ## Usage
 
+### Plugin
+
+**In order to further improve your production efficiency, we have also prepared plug-ins for you**
+
+- provider_template：[Android Studio/Intellij   Plugin](https://plugins.jetbrains.com/plugin/16821-flutter-provider)
+  - one-click generation of necessary folders, files, template codes, etc. for each page
+  - support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
```
rrousselGit, line 87:
```suggestion
  - support Wrap Widget: Consumer, Selector, ChangeNotifierProvider
```

README.md
CNAD666: The essence of create function  is to obtain the address of an object, and then store it in _InheritedProviderScopeElement

I think the provider should not be understood as a variable, it represents the address passed to the create function, and then _InheritedProviderScopeElement stores this address, which is the same as directly creating an object.

Because the provider is a final type, there is no need to worry about the provider address being changed

In this way, we can use provider externally instead of Provider.of, which is a bit troublesome.

README.md
CNAD666: Hi，this is function of The plugin I wrote，It can easily generate widget refresh control

look

- pulgin
image

- support Wrap Widget: Consumer, Selector, ChangeNotifirProvider
image

- Support input **provider** prompt to generate quick code snippets
image

README.md
CNAD666: - For those who do not understand the principle of Provider, write the following code, and it will find that Provider.of cannot trigger increase(). This is a wrong use of context, and CounterProvider cannot be found.

```dart
class CounterPage extends StatelessWidget {
  @override
    return ChangeNotifierProvider(
      create: (BuildContext context) => CounterProvider(),
      child: Scaffold(
        appBar: AppBar(title: Text('Provider')),
        body: Center(
          child: Consumer<CounterProvider>(
            builder: (context, provider, child) {
              return Text('count ${provider.count} ');
            },
          ),
        ),
        floatingActionButton: FloatingActionButton(
          onPressed: () {
            Provider.of<CounterProvider>(context, listen: false).increment();
          },
          child: Icon(Icons.add),
        ),
      ),
    );
  }
}

class CounterProvider extends ChangeNotifier {
  int count = 0;

  void increment() {
    count++;
    notifyListeners();
  }
}
```

- If we cache CounterProvider() first and then use it, we can avoid this problem

```dart
class CounterPage extends StatelessWidget {
  final provider = CounterProvider();

  @override
    return ChangeNotifierProvider(
      create: (BuildContext context) => provider,
      child: Scaffold(
        appBar: AppBar(title: Text('Provider')),
        body: Center(
          child: Consumer<CounterProvider>(
            builder: (context, provider, child) {
              return Text('count ${provider.count} ');
            },
          ),
        ),
        floatingActionButton: FloatingActionButton(
          onPressed: () => provider.increment(),
          child: Icon(Icons.add),
        ),
      ),
    );
  }
}

class CounterProvider extends ChangeNotifier {
  int count = 0;

  void increment() {
    count++;
    notifyListeners();
  }
}
```

rrousselGit: This isn't the correct solution. Instead do:

```dart
ChangeNotifierProvider<Model>(
  create: ...,
  builder: (context, child) {
    return Button(
     onPressed: () => context.read<Model>() // works
   );
  }
)

rrousselGit: If you don't do that, you can have problems if the widget rebuilds, as you'd lose your reference to the notifier

README.md
rrousselGit: But what is the "Some usage" part about? It doesn't seem related to the plugin

CNAD666: image
image


It can generate some code, you can choose to use it or not

README.md
CNAD666: > If you don't do that, you can have problems if the widget rebuilds, as you'd lose your reference to the notifier

create is just a copy of the object we passed in, stored in InheritedElement, I can also store a copy externally.

README.md
rrousselGit: But what is the relationship with the doc and this?

README.md
rrousselGit: This still is ani-pattern. 

Your application will break if somehow the widget rebuilds.

CNAD666: > This still is ani-pattern.
> 
> Your application will break if somehow the widget rebuilds.

your are right

PR #653, opened 2021-06-25
Fixes for null safety

### LINKED ISSUE #652: A couple of null safety tweaks
Hi Remi,

We're looking at rolling the latest `package:provider` into google and hit a couple of blocking issues.

We have also taken the time to investigate and have a proposed solution ;)

Both issues only show up with mixed null safe and legacy code.

1. Legacy inherited providers can't contain nulls any more. This is because of the use of `_value is! T` to check for failed initialization. Legacy code can't supply an explicitly nullable `T`, so `null` is never `T`. Proposed fix: use an additional `bool` to track whether initialization succeeded: https://github.com/davidmorgan/provider/commit/1d6ee87cd48da981e5fb9302a831ec7b56b9bcd0

2. Legacy code might want to inject a `Provider<T?>`, because legacy code is used to handling nulls. But, legacy code can't write `T?` so it can't supply the right key--it can only ever request non-nullable providers. Proposed fix: when running in unsound (mixed) mode, look for `Provider<T?>` if `Provider<T>` is not found: https://github.com/davidmorgan/provider/commit/1a55c485cba486480e86e686fde48878410c24fb

Both proposed fixes include test coverage; adding test coverage is a bit awkward because it requires tests that run in mixed mode, which needs a bit of care when running `flutter test`; I update `flutter_test.sh` to do that.

Please let me know if this seems a good direction, if so I will clean up the code and send PRs. Or, of course, happy to discuss :)

Thanks :)

davidmorgan:
Fixes #652

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -328,6 +328,17 @@ If you want to expose a variable that can be anything, consider changing
           _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>?;
     }
 
+    // When running in mixed mode, a Provider<T?> is hard to access from legacy
+    // code because it cannot write T?. Fall back from looking for Provider<T>
+    // to looking for Provider<T?> to support this case.
+    //
+    // The fallback is not needed in sound mode, where all code must be correct
+    // about nullability of types.
+    if (!_isSoundMode) {
+      inheritedElement ??= context.getElementForInheritedWidgetOfExactType<
+          _InheritedProviderScope<T?>>() as _InheritedProviderScopeElement<T>?;
+    }
+
     if (inheritedElement == null) {
       throw ProviderNotFoundException(T, context.widget.runtimeType);
     }
@@ -649,3 +660,13 @@ extension WatchContext on BuildContext {
 ///
 /// This function
 typedef Locator = T Function<T>();
+
+/// Whether the runtime has null safe sound mode enabled.
+///
+/// In sound mode, all code is null safe and null safety is enforced everywhere.
+/// Nullability in generics is also enforced, which is how this code detects
+/// sound mode.
+///
+/// In unsound mode, there can be a mix of null safe and legacy code. Some null
+/// checks are not done, and generics are not compared for null safety.
+final bool _isSoundMode = <int?>[] is! List<int>;
--- a/test/inherited_provider_legacy_test.dart
+++ b/test/inherited_provider_legacy_test.dart
@@ -1,0 +1,35 @@
+// Mixed mode: test is legacy, runtime is legacy, package:provider is null safe.
+// @dart=2.11
+import 'package:flutter/widgets.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:provider/provider.dart';
+
+void main() {
+  testWidgets('allows nulls in mixed mode', (tester) async {
+    // ignore: avoid_returning_null
+    int initialValueBuilder(BuildContext _) => null;
+
+    await tester.pumpWidget(
+      InheritedProvider<int>(
+        create: initialValueBuilder,
+        child: const Context(),
+      ),
+    );
+
+    // Prior to fix: first get initializes to null, second get throws
+    // because of the null. After fix, the null counts as valid.
+    expect(Provider.of<int>(context, listen: false), isNull);
+    expect(Provider.of<int>(context, listen: false), isNull);
+  });
+}
+
+BuildContext get context => find.byType(Context).evaluate().single;
+
+class Context extends StatelessWidget {
+  const Context({Key key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Container();
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 - Bug fix: `InheritedProvider<T>` can again provide nulls in legacy or mixed
   legacy and null safe code.
+- Support incremental migration to null safety: in unsound mode, looking up
+  `Provider<T>` will now fall back to looking for `Provider<T?>`. This allows
+  legacy code to continue to inject providers that might provide null.
 
 # 5.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# 5.0.1 (unreleased)
+
+- Bug fix: `InheritedProvider<T>` can again provide nulls in legacy or mixed
+  legacy and null safe code.
+
 # 5.0.0
 
 - Stable, null-safe release.
--- a/lib/src/inherited_provider.dart
+++ b/lib/src/inherited_provider.dart
@@ -651,13 +651,14 @@ bool debugIsInInheritedProviderCreate = false;
 class _CreateInheritedProviderState<T>
     extends _DelegateState<T, _CreateInheritedProvider<T>> {
   VoidCallback? _removeListener;
-  bool _didInitValue = false;
+  bool _triedInitValue = false;
+  bool _succeededInitValue = false;
   T? _value;
   _CreateInheritedProvider<T>? _previousWidget;
 
   @override
   T get value {
-    if (_didInitValue && _value is! T) {
+    if (_triedInitValue && !_succeededInitValue) {
       throw StateError(
         'Tried to read a provider that threw during the creation of its value.\n'
         'The exception occurred during the creation of type $T.',
@@ -674,8 +675,8 @@ class _CreateInheritedProviderState<T>
       return true;
     }());
 
-    if (!_didInitValue) {
-      _didInitValue = true;
+    if (!_triedInitValue) {
+      _triedInitValue = true;
       if (delegate.create != null) {
         assert(debugSetInheritedLock(true));
         try {
@@ -685,6 +686,7 @@ class _CreateInheritedProviderState<T>
             return true;
           }());
           _value = delegate.create!(element!);
+          _succeededInitValue = true;
         } finally {
           assert(() {
             debugIsInInheritedProviderCreate =
@@ -709,6 +711,7 @@ class _CreateInheritedProviderState<T>
             return true;
           }());
           _value = delegate.update!(element!, _value);
+          _succeededInitValue = true;
         } finally {
           assert(() {
             debugIsInInheritedProviderCreate =
@@ -724,6 +727,7 @@ class _CreateInheritedProviderState<T>
           return true;
         }());
       }
+      _succeededInitValue = true;
     }
 
     element!._isNotifyDependentsEnabled = false;
@@ -737,7 +741,7 @@ class _CreateInheritedProviderState<T>
   void dispose() {
     super.dispose();
     _removeListener?.call();
-    if (_didInitValue) {
+    if (_triedInitValue) {
       delegate.dispose?.call(element!, _value as T);
     }
   }
@@ -745,7 +749,7 @@ class _CreateInheritedProviderState<T>
   @override
   void debugFillProperties(DiagnosticPropertiesBuilder properties) {
     super.debugFillProperties(properties);
-    if (_didInitValue) {
+    if (_triedInitValue) {
       properties
         ..add(DiagnosticsProperty('value', value))
         ..add(
@@ -774,7 +778,7 @@ class _CreateInheritedProviderState<T>
     // Don't call `update` unless the build was triggered from `updated`/`didChangeDependencies`
     // otherwise `markNeedsNotifyDependents` will trigger unnecessary `update` calls
     if (isBuildFromExternalSources &&
-        _didInitValue &&
+        _triedInitValue &&
         delegate.update != null) {
       final previousValue = _value;
 
@@ -834,7 +838,7 @@ class _CreateInheritedProviderState<T>
   }
 
   @override
-  bool get hasValue => _didInitValue;
+  bool get hasValue => _triedInitValue;
 }
 
 class _ValueInheritedProvider<T> extends _Delegate<T> {
--- a/scripts/flutter_test.sh
+++ b/scripts/flutter_test.sh
@@ -3,6 +3,16 @@ cd $1
 flutter packages get
 flutter format --set-exit-if-changed lib test
 flutter analyze --no-current-package lib test/
-flutter test --no-pub --coverage
+
+# "flutter test" fails if there is a mix of null safe tests and legacy tests
+# marked with language version comment. So, run in two batches.
+
+# Null safe tests: without "_legacy_" in the name.
+flutter test --no-pub --coverage $(ls test/*_test.dart | grep -v _legacy_)
+
+# Legacy tests: with "_legacy_" in the name.
+flutter test --no-pub --coverage --no-sound-null-safety $(ls test/*_legacy_*)
+
+
 # resets to the original state
 cd -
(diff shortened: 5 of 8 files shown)


davidmorgan:
Oh, I just noticed you're also the author of `freezed`. Cool! I guess you have heard of my `built_value` ;)

rrousselGit:
> I guess you have heard of my `built_value` ;)

Indeed! :smile:

test/inherited_provider_legacy_test.dart
```diff
@@ -0,0 +1,35 @@
...
+        create: initialValueBuilder,
+        child: const Context(),
+      ),
+    );
+
+    // Prior to fix: first get initializes to null, second get throws
+    // because of the null. After fix, the null counts as valid.
+    expect(Provider.of<int>(context, listen: false), isNull);
+    expect(Provider.of<int>(context, listen: false), isNull);
+  });
```
rrousselGit, line 23: The test probably should have a `skip: isSound`. And same for the sound variant

lib/src/provider.dart
```diff
@@ -328,6 +328,17 @@ If you want to expose a variable that can be anything, consider changing
           _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>?;
     }
 
+    // When running in mixed mode, a Provider<T?> is hard to access from legacy
+    // code because it cannot write T?. Fall back from looking for Provider<T>
+    // to looking for Provider<T?> to support this case.
+    //
+    // The fallback is not needed in sound mode, where all code must be correct
+    // about nullability of types.
+    if (!_isSoundMode) {
```
rrousselGit, line 337: What do you think about having this behavior in sound mode too?

Thinking about it, I'm not a fan of having the code behave differently in sound vs unsound code. That could surprise folks as they fully migrate their app.


Or we could have the `ProviderNotFoundException` error message say "tried to read <int> but only a <int?> found. Did you want to do Provider.of<int?> instead?"

davidmorgan: This is a subtle point so I'm glad you are asking about it, in case my reasoning is not 100% correct :)

In sound mode, generics are checked for nullability. So even if the fallback succeeded and found a `Provider<T?>`, it would be impossible to return it:

```
List<int> getList() {
  return <int?>[1, 2, 3] as dynamic; // throws at runtime
}
```

even if we _did_ somehow force a `Provider<T?>` into a `Provider<T>`, it would be unable to return `null`; that would similarly be a runtime error.

Neither of those checks runs in unsound mode. In unsound mode, it's possible to treat a `Provider<T?>` as a `Provider<T>`, and it's possible for it to return `null`.

This is a feature, not a bug, because of unmigrated code. In unmigrated code, `Provider<T>` has always been able to return null, and that needs to keep working until all the code is migrated. So, unmigrated code should be able to ask for a `Provider<T>` and receive either a `Provider<T>` or a ` Provider<T?>`, depending on what is installed; from the point of view of the unmigrated code they are the same.

Re: different behaviour in unsound and sound mode: I believe this is consistent with other differences people will encounter. There are various ways you can get the nullability of generics wrong that won't show up until you enable sound mode, at which point they become runtime errors. e.g. you might be passing around a `List<int?>` as a `List<int>`, and that will work fine until you turn on sound mode. Then, you'll be forced to either make the instance a `List<int>` or refer to it as a `List<int?>`. The equivalent for providers is that you might install a `Provider<int?>` and look it up as a `Provider<int>`, as this change explicitly allows. When you enable sound mode you will have to either make it a `Provider<int>` or look it up as a `Provider<int?>`, with very similar knock-on effects to the `List` case. In practice so far we've found that the static checks that get added as you migrate are sufficient--as intended!--to limit the unpleasant surprises when you turn on sound mode to a manageable number.

We can imagine a solution whereby we could allow `Provider<int?>` to be referred to as `Provider<int>` only in legacy code, but I don't think there is a way to do it. The fact that the language does not try to do that for `List` is, I think, a pretty strong indication that this is the best we can do :)

Re: updating `ProviderNotFoundException`, the problem is that there is no way to specify `Provider<T?>` in unmigrated code, so they can't ask for it. (Not 100% true, they could call migrated code to do so like I do in the test--but this would be unwieldy and a breaking change).

Happy to discuss further, as I said, a tricky point :) thanks.

test/inherited_provider_legacy_test.dart
davidmorgan: (thanks, will address after the other discussion is done)

lib/src/provider.dart
rrousselGit: > In sound mode, generics are checked for nullability. So even if the fallback succeeded and found a Provider<T?>, it would be impossible to return it:

I don't think that's really an issue. Something I had in mind to improve the null-safety experience was:

- if `T` is non-nullable and a provider is not found, throw ProviderNotFoundException
- if `T` is nullable, then when a provider is not found, return `null`

So we could have:

```dart
of<T>() {
  // providers' generic is updated to always be nullable
  InheritedElement? element = context.dependOn<Provider<T?>>();

  if (element == null && null is! T) throw ProviderNotFoundException(...);

  T? value = element?.value;
  if (value == null && null is! T) throw ProviderNotFoundException(...);
  return value;
}
```

This both fixes the issue of unsound vs sound code and allows people to support cases where a provider does not exist. Like:

```dart
runApp(
  Consumer<Movie?>(
    builder: (context, value, _) {
      if (value == null) return Text('no movie found');
      return Text(movie.title);
    },
  ),
)
```

> Re: updating ProviderNotFoundException, the problem is that there is no way to specify Provider<T?> in unmigrated code, so they can't ask for it. (Not 100% true, they could call migrated code to do so like I do in the test--but this would be unwieldy and a breaking change).


That sounds like the true issue with unmigrated code.

Maybe an alternative is to have provider expose the following?

```dart
class Provider {
  static legacyOf<T>(context) {
    return of<T?>(context);
  }
}
```

This would allow unmigrated code to do `Provider.of<T?>` without having to change the behavior of migrated code.

Would that work? That could be a simpler way to unblock potential issues.

davidmorgan: > > In sound mode, generics are checked for nullability. So even if the fallback succeeded and found a Provider<T?>, it would be impossible to return it:
> 
> I don't think that's really an issue. Something I had in mind to improve the null-safety experience was:
> 
> * if `T` is non-nullable and a provider is not found, throw ProviderNotFoundException
> * if `T` is nullable, then when a provider is not found, return `null`
> 
> So we could have:
> 
> ```dart
> of<T>() {
>   // providers' generic is updated to always be nullable
>   InheritedElement? element = context.dependOn<Provider<T?>>();
> 
>   if (element == null && null is! T) throw ProviderNotFoundException(...);
> 
>   T? value = element?.value;
>   if (value == null && null is! T) throw ProviderNotFoundException(...);
>   return value;
> }
> ```
> 
> This both fixes the issue of unsound vs sound code and allows people to support cases where a provider does not exist. Like:
> 
> ```dart
> runApp(
>   Consumer<Movie?>(
>     builder: (context, value, _) {
>       if (value == null) return Text('no movie found');
>       return Text(movie.title);
>     },
>   ),
> )
> ```

Interesting.

By always looking up with `T?`, you lose the ability to register a separate provider for `T` and for `T?`. That sounds better to me, having both seems likely to be confusing. If you really want both you can have e.g. `Provider<Optional<T>>`.

FWIW I feel like defaulting to providing null from no provider is undesirable; `T` is usually specified statically, so having a missing provider for `T` seems most likely to be a programmer error and throwing would be better. Or maybe a different method for the permissive way, `tryOf`? But, that's a design decision and orthogonal to the discussion here.

Back on topic: unfortunately, the implementation suffers from the same problem, which is that legacy code can't easily express `T?` and so `of` will unexpected throw on `null` when called from legacy code. I think the possible fixes are the same: either run the check in sound mode only, or force legacy code to jump through hoops to pass `T?`. Since that is your next suggestion, let us go on to that one :)

> > Re: updating ProviderNotFoundException, the problem is that there is no way to specify Provider<T?> in unmigrated code, so they can't ask for it. (Not 100% true, they could call migrated code to do so like I do in the test--but this would be unwieldy and a breaking change).
> 
> That sounds like the true issue with unmigrated code.
> 
> Maybe an alternative is to have provider expose the following?
> 
> ```dart
> class Provider {
>   static legacyOf<T>(context) {
>     return of<T?>(context);
>   }
> }
> ```
> 
> This would allow unmigrated code to do `Provider.of<T?>` without having to change the behavior of migrated code.
> 
> Would that work? That could be a simpler way to unblock potential issues.

For google that's problematic because of our monorepo--we can't change `of` to `legacyOf` across the repo as we roll in the new `provider`, there are too many, so we'd have to do it in stages with a locally patched `provider` to add the new API before the move to null safety. More likely we would instead patch in the runtime behaviour proposed in this PR, as that should give almost as good results with less churn. It's true that it would give less runtime checking while migration is in progress; but the runtime checks are usually of secondary importance to the static checks, i.e. that once a `Provider<T>` is migrated, it's much less likely to return a null due to the static checks. (We require that migrated code depend only on migrated code, so the only way to get a runtime violation of the static checks is to inject a badly behaved `Foo<T>` from legacy code into migrated code).

The intention of the null safety migration is that such breaking API changes aren't needed, and that this is achieved by having partial null safety migration be static only and local only. Static checks spread through the code as you incrementally migrate, but the full runtime checks only arrive with sound mode. Of course, nothing says that everyone has to follow this pattern :) for the most part I think most developers don't have to be aware of the subtleties here.

Next steps:

--> should I split the `InheritedProvider` fix into a separate PR, as that seems uncontentious?
--> please let me know if you want me to implement `T`->`T?` mapping on lookup (and presumably we need the same when registering a provider). This seems a reasonable way to go.
--> up to you to decide on the runtime behaviour for unsound mode :)

Thanks :)

rrousselGit: > --> should I split the InheritedProvider fix into a separate PR, as that seems uncontentious?

That sound good  :+1: 

> --> please let me know if you want me to implement T->T? mapping on lookup (and presumably we need the same when registering a provider). This seems a reasonable way to go.

If you want to have a go at it, sure :smile:
Otherwise, I can work on it in a few days.

davidmorgan: Split out a PR #654 

Re: `T`->`T?`, implementation looks problematic: the registration happens deep in Flutter `framework.dart` here:

```dart
  @override
  void _updateInheritance() {
      ...
    _inheritedWidgets![widget.runtimeType] = this;
  }
```

which as far as I can see means to make this work we have to override `runtimeType` on `_InheritedProviderScope`:

```
  @override
  Type get runtimeType => _typeOf<_InheritedProviderScope<T?>>();
  static Type _typeOf<X>() => X;
```

which "works", in that most tests pass, but is likely to cause other horrible breakage; e.g. some test expectations fail because the class now lies about its type

```
  Expected: contains 'InheritedProvider<int>(value: <not yet loaded>)'
  Actual: 'InheritedProvider<int?>(value: <not yet loaded>)'
```

As far as I can see that rules out the 'always look up T?' approach.

Taking a step back, I think the goal was to prevent confusion between `Provider<T>` and `Provider<T?>` by only allowing one to be installed and using it for both cases. Things we can do:

 - On lookup, if given non-nullable `T`, we can try to look up both `Provider<T>` and `Provider<T?>`
 - Oddly I think this is asymmetrical: if given nullable `T` it is much harder to look up any non-nullable `Provider<T>`, I think you would have to build a `Map<Type, Type>` from nullable to non-nullable types; you could populate the map when each non-nullable `Provider<T>` is instantiated. I'll check with the language team to see if there's another way...
 - On adding a `Provider<T>` we could check if `Provider<T?>` is already installed and make that an error; the other way around is similarly more difficult and needs the map
 - Another path: we could prohibit `Provider<T?>` by declaring `T extends Object`, which reduces the confusion but also breaks all functionality around nulls. Maybe a different API for nulls? Unfortunately that's a breaking change.

If you have additional ideas please that'd be great :) I'll update when I get a reply from the language team.

rrousselGit: > which as far as I can see means to make this work we have to override `runtimeType` on `_InheritedProviderScope`:

This isn't necessary.

Instead what can be done is editing where `_InheritedProviderScope` is instantiated, as the class is private and the package controls all places where it is used.

So `InheritedProvider` and `DeferredInheritedProvider`'s `buildWithChild` method could be changed like so:

```diff
-    return _InheritedProviderScope<T>(
+    return _InheritedProviderScope<T?>(
```

davidmorgan: Ah, I didn't spot that--thanks.

Still not looking easy I'm afraid--looks like that change has knock on effects elsewhere

```
The following _TypeError was thrown building Provider<int>(value: 0):
type '_InheritedProviderScopeElement<int?>' is not a subtype of type
'_InheritedProviderScopeElement<int>?' of 'value'

The relevant error-causing widget was:
  Provider<int>
  file:///usr/local/google/home/davidmorgan/git/provider/test/inherited_provider_test.dart:2330:20

When the exception was thrown, this was the stack:
#0      _DelegateState.element= (package:provider/src/inherited_provider.dart)
#1      _InheritedProviderScopeElement.performRebuild (package:provider/src/inherited_provider.dart:454:62)
```

--will need to follow that trail and see if it can be fixed without changes to the public API.

rrousselGit: The failing test can be changed. The issue is the test itself, not the change  😄

davidmorgan: Unfortunately it's not just the test; changing the type from `T` to `T?` so we have a better key breaks the implementation.

The problem is that `T?` is going to leak through everywhere--as it should!--to require everywhere in the implementation to accept/handle nulls always. In the end, it will leak through to the public API, so that a `Provider<T>` returns a value of type `T?`. I think the only way we could avoid that would be to parameterize internally by two types, `T` and `NT`, where `NT` is the nullable `T`. But that defeats the who point, because raw `T` is now part of the key type again, and we have the lookup problem again.

The language team responded in the negative to my question: there is no trivial way to get from `T?` to `T` on lookup.

I created a spreadsheet with my understanding of the situation and the changes that seem workable so far:

https://docs.google.com/spreadsheets/d/1vPDhL7sZKqGabLZ0KspvSKZZ-1oySKtCUaSLkajb-pA/edit#gid=0

While enumerating the cases I realize there is a converse problem to the one this PR was opened for: if you have an unmigrated `Provider<T*>` then you can't inject it as a `Provider<T?>`. So even with this change, legacy providers that might return null must migrate before the injection sites.

I'm happy to keep discussing/working towards a solution here, although I'm currently out of ideas for any way to improve on this PR as originally suggested. (Don't see a way to implement any of the suggestions that followed). If you think it would be helpful I'm happy to chat in realtime or VC, please let me know :)

Thanks.

rrousselGit: A v6.0.0 version is likely necessary in any case, as some users may have both `T` and `T?`. 

> In the end, it will leak through to the public API

I don't believe that it will, no.
`Provider<T>` will not return `T?` because `Provider.of` will handle potential mismatches.


It may be easier for me to open a PR with the changes I have in mind and have you take a look. Is that fine with you? :smile:

davidmorgan: Yes please, that sounds like a good next step :) thanks.

rrousselGit: I have opened a PR with the proposed changes: https://github.com/rrousselGit/provider/pull/655

rrousselGit:
Closing as a separate PR for this was merged

PR #654, opened 2021-06-25
Minimal repro and fix for mixed mode issue.

davidmorgan:
Fix for first issue in #652 

In the other PR review you suggested tagging a test that only runs in unsound mode with `skip` for sound mode; but I think that isn't needed here. Because of the `// @dart=2.11` comment, the test will anyway refuse to run at all in sound mode, as language version >=2.12 everywhere is required for sound mode. (The other side of your suggestion does make sense, and is for the other PR: tagging a sound mode only test so that it will not run in unsound mode).

Thanks.


rrousselGit (approved):
Looking good, thanks!

PR #655, opened 2021-06-28
Support conditionally depending on a provider

### LINKED ISSUE #652: A couple of null safety tweaks

rrousselGit:
Fixes #652

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -17,6 +17,16 @@ part 'deferred_inherited_provider.dart';
 part 'devtool.dart';
 part 'inherited_provider.dart';
 
+/// Whether the runtime has null safe sound mode enabled.
+///
+/// In sound mode, all code is null safe and null safety is enforced everywhere.
+/// Nullability in generics is also enforced, which is how this code detects
+/// sound mode.
+///
+/// In unsound mode, there can be a mix of null safe and legacy code. Some null
+/// checks are not done, and generics are not compared for null safety.
+final bool _isSoundMode = <int?>[] is! List<int>;
+
 /// A provider that merges multiple providers into a single linear widget tree.
 /// It is used to improve readability and reduce boilerplate code of having to
 /// nest multiple layers of providers.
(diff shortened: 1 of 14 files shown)


rrousselGit:
cc @davidmorgan

lib/src/provider.dart
```diff
@@ -314,26 +338,27 @@ If you want to expose a variable that can be anything, consider changing
...
         return false;
       });
     } else {
       inheritedElement = context.getElementForInheritedWidgetOfExactType<
-          _InheritedProviderScope<T>>() as _InheritedProviderScopeElement<T>?;
+          _InheritedProviderScope<T?>>() as _InheritedProviderScopeElement<T?>?;
     }
 
-    if (inheritedElement == null) {
+    if (inheritedElement == null && null is! T) {
```
davidmorgan, line 357: This one too, I think this needs guarding with `_isSoundMode`.

davidmorgan (commented):
Thanks!

Great, it does look like this approach works--the impact of `T?` is strictly speaking visible in the public API but it looks like a rarely used part, so impact is minimal :) i.e. it's effectively hidden behind `Provider`. (As you said!)

The sheer number of cases to cover for full test coverage is intimidating, in case it's useful, I added them all for basic InheritedProvider inject here (but didn't tidy up the code): https://github.com/rrousselGit/provider/commit/0afbd42d9689e7c9cf07767703d465a62ce366d5 and those all work as expected/desired.

In the `of` check you guard the `value is! T` check with  `_isSoundMode`, which is what lets mixed mode continue to pass around nulls as before, allowing an incremental migration--I've commented on other such checks which should probably be guarded in the same way, although I did not take the time to create test cases so I'm not 100% sure :)

Thanks for looking into this--please let me know how I can help with next steps :)

davidmorgan:
@rrousselGit Anything I can do to help move this forward, please? Not sure if you wanted me to adjust my PR in this direction or if you were planning on taking/landing it. Thanks.

davidmorgan:
Hi hi -- any update please? :) No specific urgency from my side, but we definitely need this at some point. Thanks.

davidmorgan:
Any update please?... Mostly to check that you're not waiting for me to do something :) thanks.

rrousselGit:
Oh I'm sorry, I didn't receive a notification.

I plan to finish and release this feature this week-end

davidmorgan:
@rrousselGit No worries--sounds good--looking forward to it :) thanks!

rrousselGit:
@davidmorgan this was published in a dev release: 6.0.0-dev

If everything is alright, I will publish a stable release in a week or two

davidmorgan:
@rrousselGit Thanks, much appreciated--I was already running our regression tests with it, all green. I will additionally verify that this unblocks incremental migration as hoped, and get back to you by Wed. I'm being specific about Wednesday as I'll be on vacation for a week from Thursday :)

Thanks again.

davidmorgan:
@rrousselGit Confirmed it works as planned :) thanks!

PR #693, opened 2021-10-28
exposed dispose to FutureProvider

iulian0512:


rrousselGit:
Hello!

Could you add tests, and showcase how it'd be used?
Because it isn't quite clear for me how that would be useful, since `dispose` wouldn't have access to what is inside the `create` call

rrousselGit (changes requested):
Hello!

Could you add tests, and showcase how it'd be used?
Because it isn't quite clear for me how that would be useful, since dispose wouldn't have access to what is inside the create call

rrousselGit:
Hello!

Do you still plan to work on this?

iulian0512:
yes, but did not had the time to do it. you can close this pr then I will reopen, when I have the time.

rrousselGit:
Perfect thanks!

PR #707, opened 2021-12-16
Explicitly mention rebuild in context.select() doc

Lootwig:
The current description might make it seem like the `context.select()` extension is useless without an explicit rebuild (which a StatelessWidget can't do through setState()). I tried to make it more obvious that this is actually taken care of by the package.


rrousselGit:
Perfect thanks!

Issue #705, opened 2021-12-07
Cannot find stack trace for exception in model constructor


jmewes:
**Describe the bug**

If there is an exception in the model constructor, then the message and stack trace of that exception cannot be found in the message and stack trace printed to the console.

Here is the start of what is actually printed:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞═══════════════════════════════════════════════════════════
The following StateError was thrown building Provider<ModelWithBrokenConstructor>:
Bad state: Tried to read a provider that threw during the creation of its value.
The exception occurred during the creation of type ModelWithBrokenConstructor.

The relevant error-causing widget was:
  Provider<ModelWithBrokenConstructor>
  Provider:file:///home/janux/src/misc/provider/test/provider_test.dart:200:11

When the exception was thrown, this was the stack:
#0      _CreateInheritedProviderState.value (package:provider/src/inherited_provider.dart:701:7)
#1      _CreateInheritedProviderState.debugFillProperties (package:provider/src/inherited_provider.dart:793:44)
```

See [console-output.txt](https://github.com/rrousselGit/provider/files/7669905/console-output.txt) for details.

**To Reproduce**

Use a provider with a model which throws an exception in its constructor.

```dart
testWidgets('includes constructor error in stack trace', (tester) async {
  try {
    await tester.pumpWidget(
      Provider(
        lazy: false,
        create: (_) => ModelWithBrokenConstructor(),
        child: Container(),
      ),
    );  
  } catch (e) {
    expect(e, contains('Catch me, if you can!'));
  }
});
```

See https://github.com/rrousselGit/provider/pull/704

**Screenshot**

![Selection_508](https://user-images.githubusercontent.com/5235584/145067988-d89e1567-3a61-48bd-84cb-4171d237dcb7.png)

**Expected behavior**

The message and stack trace of the exception in the model constructor should be printed to the console.

**Additional Context**

As a workaround, one can run the test or app in debug mode. Then the debugger automatically sets a breakpoint right after the throwing of the exception. So, one can see there what was the root cause of the error.

PR #711, opened 2021-12-20
fix: added error details for provider that threw during the creation

jmewes:
Solves GH-705

### DIFF
--- a/lib/src/inherited_provider.dart
+++ b/lib/src/inherited_provider.dart
@@ -757,13 +757,6 @@ class _CreateInheritedProviderState<T>
             return true;
           }());
           _value = delegate.update!(element!, _value);
-        } catch (e, stackTrace) {
-          _initError = FlutterErrorDetails(
-            library: 'provider',
-            exception: e,
-            stack: stackTrace,
-          );
-          rethrow;
         } finally {
           assert(() {
             debugIsInInheritedProviderCreate =
--- a/lib/src/inherited_provider.dart
+++ b/lib/src/inherited_provider.dart
@@ -691,18 +691,18 @@ class _CreateInheritedProviderState<T>
     extends _DelegateState<T, _CreateInheritedProvider<T>> {
   VoidCallback? _removeListener;
   bool _didInitValue = false;
-  bool _didSucceedInit = false;
   T? _value;
   _CreateInheritedProvider<T>? _previousWidget;
   FlutterErrorDetails? _initError;
 
   @override
   T get value {
-    if (_didInitValue && !_didSucceedInit) {
+    if (_didInitValue && _initError != null) {
       throw StateError(
-          'Tried to read a provider that threw during the creation of its value.\n'
-          'The exception occurred during the creation of type $T.\n\n'
-          '${_initError?.toString()}');
+        'Tried to read a provider that threw during the creation of its value.\n'
+        'The exception occurred during the creation of type $T.\n\n'
+        '${_initError?.toString()}',
+      );
     }
     bool? _debugPreviousIsInInheritedProviderCreate;
     bool? _debugPreviousIsInInheritedProviderUpdate;
@@ -726,7 +726,6 @@ class _CreateInheritedProviderState<T>
             return true;
           }());
           _value = delegate.create!(element!);
-          _didSucceedInit = true;
         } catch (e, stackTrace) {
           _initError = FlutterErrorDetails(
             library: 'provider',
@@ -758,7 +757,13 @@ class _CreateInheritedProviderState<T>
             return true;
           }());
           _value = delegate.update!(element!, _value);
-          _didSucceedInit = true;
+        } catch (e, stackTrace) {
+          _initError = FlutterErrorDetails(
+            library: 'provider',
+            exception: e,
+            stack: stackTrace,
+          );
+          rethrow;
         } finally {
           assert(() {
             debugIsInInheritedProviderCreate =
@@ -774,7 +779,6 @@ class _CreateInheritedProviderState<T>
           return true;
         }());
       }
-      _didSucceedInit = true;
     }
 
     element!._isNotifyDependentsEnabled = false;
--- a/lib/src/inherited_provider.dart
+++ b/lib/src/inherited_provider.dart
@@ -694,14 +694,15 @@ class _CreateInheritedProviderState<T>
   bool _didSucceedInit = false;
   T? _value;
   _CreateInheritedProvider<T>? _previousWidget;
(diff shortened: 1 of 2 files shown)


jmewes:
### Before

Before this PR, the `message` in `StateError` created by the test case `StateError is thrown when exception occurs in create` looked like this:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞════════════════════════
The following StateError was thrown building Provider<String>:
Bad state: Tried to read a provider that threw during the
creation of its value.
The exception occurred during the creation of type String.

The relevant error-causing widget was:
  Provider<String>
  Provider:file:///home/janux/src/misc/provider/test/inherited_provider_test.dart:2451:7

When the exception was thrown, this was the stack:
#0      _CreateInheritedProviderState.value (package:provider/src/inherited_provider.dart:706:7)
...
<asynchronous suspension>
<asynchronous suspension>
(elided 5 frames from dart:async and package:stack_trace)

═════════════════════════════════════════════════════════════════
```

Details: [error_before.txt](https://github.com/jmewes/provider/files/7744900/error_before.txt)

### After

After this PR, the `message` in `StateError` created by the test case `StateError is thrown when exception occurs in create` looks like this:

```
══╡ EXCEPTION CAUGHT BY WIDGETS LIBRARY ╞════════════════════════
The following StateError was thrown building Provider<String>:
Bad state: Tried to read a provider that threw during the
creation of its value.
The exception occurred during the creation of type String.

══╡ EXCEPTION CAUGHT BY PROVIDER ╞═══════════════════════════════
The following _Exception was thrown:
Exception: oops

When the exception was thrown, this was the stack:
#0      main.<anonymous closure>.<anonymous closure>
...
#40     testWidgets.<anonymous closure>.<anonymous closure>
(package:flutter_test/src/widget_tester.dart:170:29)
<asynchronous suspension>
<asynchronous suspension>
(elided 5 frames from dart:async and package:stack_trace)
═════════════════════════════════════════════════════════════════

The relevant error-causing widget was:
  Provider<String>
  Provider:file:///home/janux/src/misc/provider/test/inherited_provider_test.dart:2451:7

When the exception was thrown, this was the stack:
#0      _CreateInheritedProviderState.value (package:provider/src/inherited_provider.dart:702:7)
...
#51     testWidgets.<anonymous closure>.<anonymous closure> (package:flutter_test/src/widget_tester.dart:170:29)
<asynchronous suspension>
<asynchronous suspension>
(elided 5 frames from dart:async and package:stack_trace)

═════════════════════════════════════════════════════════════════

```

Details: [error_after.txt](https://github.com/jmewes/provider/files/7744903/error_after.txt)

lib/src/inherited_provider.dart
```diff
@@ -694,14 +694,15 @@ class _CreateInheritedProviderState<T>
   bool _didSucceedInit = false;
```
rrousselGit, line 694: If we're storing the error/stack, this can be removed

lib/src/inherited_provider.dart
```diff
@@ -694,14 +694,15 @@ class _CreateInheritedProviderState<T>
...
   @override
   T get value {
     if (_didInitValue && !_didSucceedInit) {
       throw StateError(
-        'Tried to read a provider that threw during the creation of its value.\n'
-        'The exception occurred during the creation of type $T.',
-      );
+          'Tried to read a provider that threw during the creation of its value.\n'
+          'The exception occurred during the creation of type $T.\n\n'
+          '${_initError?.toString()}');
```
rrousselGit, line 705: nit: trailing comma

lib/src/inherited_provider.dart
```diff
@@ -694,14 +694,15 @@ class _CreateInheritedProviderState<T>
   bool _didSucceedInit = false;
   T? _value;
   _CreateInheritedProvider<T>? _previousWidget;
+  FlutterErrorDetails? _initError;
 
   @override
   T get value {
     if (_didInitValue && !_didSucceedInit) {
       throw StateError(
```
rrousselGit, line 702: Now that rethrowing an error with a specific stacktrace is a thing (Error.throwWithStackTrace), maybe we should replacd the StateError in favor of throwing the original error?

rrousselGit:
Thanks! I've left a few comments. Nothing major

lib/src/inherited_provider.dart
jmewes: The currents tests are explicitly validating that there is a state error:

https://github.com/rrousselGit/provider/blob/0bbe2269a79e3c50e611cb9b29d38cf120dee77a/test/inherited_provider_test.dart#L2406

I don't know if it would make sense to catch the `StateError` somewhere, but if somebody does, this would be a breaking change.

----

I have looked up [`Error.throwWithStackTrace`](https://api.dart.dev/dev/2.16.0-9.0.dev/dart-core/Error/throwWithStackTrace.html) and it turns out that this method will be introduced with Dart version 2.16. The Flutter stable channel is currently using Dart version 2.15.1. The Flutter dev channel is using Dart version 2.16.0-80.1.beta.

----------

So, if you ask for my opinion, I would stick to throwing the `StateError`, at least until the Flutter stable channel uses Dart 2.16.

Even then, it would be a rather cosmetic change for a rather unusual edge case. So, even though I think that `Error.throwWithStackTrace` may be more elegant, I personally would not do this change, just because of the required effort.

----------

If I may rant a little, I am very surprise that Dart exceptions do not have a `cause` parameter for exceptions, like I am used to from Java. For me the best solution would be pass the `_initError` as `cause` parameter into the `StateError`. Then the Dart VM can figure out how to print it to the console.

lib/src/inherited_provider.dart
```diff
@@ -691,16 +691,17 @@ class _CreateInheritedProviderState<T>
...
-  bool _didSucceedInit = false;
   T? _value;
   _CreateInheritedProvider<T>? _previousWidget;
+  FlutterErrorDetails? _initError;
 
   @override
   T get value {
-    if (_didInitValue && !_didSucceedInit) {
+    if (_didInitValue && _initError != null) {
       throw StateError(
```
rrousselGit, line 701:
```suggestion
      // TODO(rrousselGit) update to use Error.throwWithStacktTrace when it reaches stable
      throw StateError(
```

lib/src/inherited_provider.dart
rrousselGit: No problem, this can wait

rrousselGit:
LGTM, thanks!

jmewes:
Thanks for taking care of the PR!

PR #729, opened 2022-01-30
Improve sentence that explains `listen` parameter usage

LSViana:
The previous sentence was reading non-stop which was making it hard to understand it during the first read.

This is a suggestion based on my experience reading the documentation, but maybe it's clear enough already and I just failed to read it properly. 😅


rrousselGit:
Lgtm, thanks!

PR #750, opened 2022-04-27
Improve performance of `_inheritedElementOf`

jiahaog:
We're looking into startup performance for a large flutter app - think deeply nested widget trees using `package:provider`. This is the specific line of interest:

https://github.com/rrousselGit/provider/blob/1fce240150fc098fc15cda61615a2fe0cea4c961/lib/src/provider.dart#L341

Zooming in on a startup CPU profile of Flutter's UI thread, we see that the `_inheritedElementOf` method takes 20ms.

image

In the Dart VM implementation, `<obj> is <type>` checks with many different types for `<obj>` like this line at runtime is relatively expensive (cc @mkustermann). This shows up as `DRT_InstanceOf` in the above profile.

This is the same view of the CPU profile after applying this PR, showing that the runtime of `_inheritedElementOf` has halved and we no longer see those calls to `DRT_InstanceOf`.

image

Please let me know if I'm mistaken, but it appears that this conditional was previously added as an optimization to avoid calling [`visitAncestorElements`](https://api.flutter.dev/flutter/widgets/BuildContext/visitAncestorElements.html). This method does takes O(n) in the depth of the tree if we go through the tree in its entirety, but since we're returning `false` in the closure to find the parent, `visitAncestorElements` would terminate quickly in constant time and the optimization is not needed (and is in fact detrimental to performance due to the cost of the `is` check).

### DIFF
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -338,19 +338,11 @@ If you want to expose a variable that can be anything, consider changing
     );
     _InheritedProviderScopeElement<T?>? inheritedElement;
 
-    if (context.widget is _InheritedProviderScope<T?>) {
-      // An InheritedProvider<T>'s update tries to obtain a parent provider of
-      // the same type.
-      context.visitAncestorElements((parent) {
-        inheritedElement = parent.getElementForInheritedWidgetOfExactType<
-                _InheritedProviderScope<T?>>()
-            as _InheritedProviderScopeElement<T?>?;
-        return false;
-      });
-    } else {
-      inheritedElement = context.getElementForInheritedWidgetOfExactType<
+    context.visitAncestorElements((parent) {
+      inheritedElement = parent.getElementForInheritedWidgetOfExactType<
           _InheritedProviderScope<T?>>() as _InheritedProviderScopeElement<T?>?;
-    }
+      return false;
+    });
 
     if (inheritedElement == null && null is! T) {
       throw ProviderNotFoundException(T, context.widget.runtimeType);
--- a/lib/src/provider.dart
+++ b/lib/src/provider.dart
@@ -336,10 +336,8 @@ If you want to expose a variable that can be anything, consider changing
 `dynamic` to `Object` instead.
 ''',
     );
-    _InheritedProviderScopeElement<T?>? inheritedElement =
-        context.getElementForInheritedWidgetOfExactType<
-                _InheritedProviderScope<T?>>()
-            as _InheritedProviderScopeElement<T?>?;
+    final inheritedElement = context.getElementForInheritedWidgetOfExactType<
+        _InheritedProviderScope<T?>>() as _InheritedProviderScopeElement<T?>?;
 
     if (inheritedElement == null && null is! T) {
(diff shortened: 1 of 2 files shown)


rrousselGit:
Hello!
Interesting!


That `is` isn't here as an optimization. It was there to make sure that we're trying to read providers from inside widgets, then we're using `context` instead of the `context`'s parent .

But it's probably safe to remove indeed.


Although thinking about it, there's probably a different solution: Remove this code entirely and only call getElementForInheritedWidgetOfExactType with no visitAncestorElement / is involved.  
And override `_InheritedProviderScope`'s `createElement` to return a custom `InheritedElement` which overrides `getElementForInheritedWidgetOfExactType` to perform that visitAncestorElement

This means that only providers would perform the visitAncestorElement, not consumers.

jiahaog:
Thanks for the suggestion! Sorry, I'm not too familiar with the internals of provider but I can try this out. I see that `_InheritedProviderScope`'s `createElement` [already returns](https://github.com/rrousselGit/provider/blob/1fce240150fc098fc15cda61615a2fe0cea4c961/lib/src/inherited_provider.dart#L354) a `_InheritedProviderScopeElement`. Do you mean that we should override `getElementForInheritedWidgetOfExactType` on `_InheritedProviderScopeElement`?

rrousselGit:
Ah I forgot that it already had a custom element

Yes, that should do.

jiahaog:
Hope I understood you correctly, updated!

lib/src/provider.dart
```diff
@@ -336,21 +336,10 @@ If you want to expose a variable that can be anything, consider changing
 `dynamic` to `Object` instead.
 ''',
     );
-    _InheritedProviderScopeElement<T?>? inheritedElement;
-
-    if (context.widget is _InheritedProviderScope<T?>) {
-      // An InheritedProvider<T>'s update tries to obtain a parent provider of
-      // the same type.
```
rrousselGit, line 343: Could you move the comment over too?

rrousselGit:
We're almost there 😊

lib/src/provider.dart
jiahaog: Done!

jiahaog:
I made the changes! Before you merge (if it looks good to you), can you let me collect a CPU profile and run the benchmark again to be sure that the new changes do have a positive effect? I'll report back as soon as I can 🙂

rrousselGit:
LGTM thanks!

Feel free to run the benchmark again

jiahaog:
Thanks for waiting and the review, I think this should be good to go now! I verified that the CPU profile after the suggested changes look similar to the ones posted in the first comment at https://github.com/rrousselGit/provider/pull/750/commits/ac108c220e39824db6e9b182aef16bf3c2514f0c.

jiahaog:
@rrousselGit friendly ping, is this good to merge now 🙂

jiahaog:
Hi @rrousselGit just checking in again :). Is there anything else needed on my end for this PR to be merged? Thanks!

rrousselGit:
Done, thanks!

jiahaog:
Thank you!

PR #751, opened 2022-04-29
Release error messages

dnfield:
By default, Flutter replaces `toString` overrides on non-`Exception` or `Error` objects (unless a pragma is specified to keep them).

These types do not get shaken out because they're exceptions, and so the very helpful messages for developers end up in release binaries (and possibly logs) for end users.

This patch guards the longer strings in release mode and replaces them with terser versions that are slightly less helpful. It marginally reduces binary size.


rrousselGit:
Looking good, thanks!

PR #759, opened 2022-06-08
Mandatory Initial Data removed

kirchesch:
Most of the times initial data is manually set to null by the user


rrousselGit:
Hello!
I suppose that since the 6.0.0, such change should be reasonable.

Could you add tests though?

Kirchesch:
Hey @rrousselGit i created a test area here: https://github.com/KircheschContributions/FlutterIssues

Just modify the main() function to run providerPrTestMain(); instead of userChangesStreamingOldDataMain();

I have tested and it is working fine, the test is plug and play, i hope this is what you meant by provide tests

rrousselGit:
No, I meant adding widget tests in the test folder of the package.

Kirchesch:
I added the tests, i did not knew how they worked, started using flutter recently, good to know about it

test/mandatory_initial_data_removal_test.dart
```diff
@@ -0,0 +1,83 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mockito/mockito.dart';
+import 'package:provider/provider.dart';
+import 'common.dart';
+
+void main() {
+  group('Test group', () {
```
rrousselGit, line 10: Could you move those tests next to where FutureProvider/StreamProvider are tested?

rrousselGit:
Hello! I apologize for the slow review. This went off my radar.

That's technically breaking, which is problematic.   
Folks could be doing `updateShouldNotify: (a, b) => a.equal(b)` without issue. Now they'd have to use `?` or `!`

For that reason alone and the fact that Provider is in long-term maintenance mode, I don't think we should go forward with this change.


I really appreciate your effort though, and apologize once again for the slow review.

PR #768, opened 2022-08-01
Implement search capability to provider.

Ayad-Ben-Saliem:
Implement search capability to provider.
A discussion opened here [https://github.com/rrousselGit/provider/discussions/737]
1- Add searchCallback to Provider.of<T>() method to give the user the ability to search for a specific value.
2- Add searchCallback to context.read<T>() and context.watch<T>() methods and pass it to Provider.of<T>() method.
3- Add SearchContext to expose context.search<T>() method.
4- Export SearchContext in /lib/provider.dart


rrousselGit:
Hello and thanks for the PR!

I'll admit that I'm not very comfortable with having this feature in package:provider. Especially considering the package probably is going to be deprecated in the future.

Is it truly necessary to have this in the official package? Could you not implement this directly in your project?

Ayad-Ben-Saliem:
Thank you for your reply!

Through Flutter projects I have worked on, I have noted this feature will be helpful if added.
Also, I'm planning to add this feature in [`Bloc`](https://github.com/felangel/bloc/tree/master/packages/flutter_bloc) package too, since it depends on the `package:Provider`.

However, this feature will not affect the previous use of `Provider`.

Another thing...
**_"Especially considering the package probably is going to be deprecated in the future."_**
Can you please tell me more? Is the package will be embedded in the Flutter framework?

rrousselGit:
Closing as I don't plan on adding support for this

PR #792, opened 2022-12-13
Make tests less fragile

goderbauer:
In https://github.com/flutter/flutter/pull/116924 Flutter is changing the shape of the root of the widget tree, which caused some of the provider tests run as part of the "Flutter Customer Test Registry" [1] to fail. These tests were making some assumptions about the root of the widget tree. This PR refactors those tests to remove that assumption, which wasn't necessary to test the specified functionality.

[1] https://github.com/flutter/tests/blob/main/registry/provider.test

### DIFF
--- a/test/inherited_provider_test.dart
+++ b/test/inherited_provider_test.dart
@@ -554,7 +554,7 @@ The context used was: Context
         ),
       );
 
-      final rootElement = tester.allElements.first;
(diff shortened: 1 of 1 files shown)


rrousselGit:
Lgtm! Thanks for making the fix 😊

PR #794, opened 2022-12-15
Use strict analysis modes instead of deprecated 'strong-mode' options

srawlins:
See https://dart.dev/guides/language/analysis-options#enabling-additional-type-checks for more info.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -2,9 +2,9 @@ include: all_lint_rules.yaml
 analyzer:
   exclude:
     - "**/*.g.dart"
-  strong-mode:
-    implicit-casts: false
-    implicit-dynamic: false
+  language:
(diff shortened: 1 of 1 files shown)


analysis_options.yaml
```diff
@@ -2,9 +2,9 @@ include: all_lint_rules.yaml
 analyzer:
   exclude:
     - "**/*.g.dart"
-  strong-mode:
-    implicit-casts: false
-    implicit-dynamic: false
+  language:
+    strict-casts: true
```
rrousselGit, line 6:
```suggestion
    strict-casts: true
    strict-inference: true
```

rrousselGit:
LGTM assuming the CI passes

srawlins:
Thanks!

PR #798, opened 2023-01-09
Change Notifier LifeCycle

muhammadusamaijaz:
Added onInit() and onReady() callbacks which will be used to initialize our late initializers in ChangeNotifiers. onReady() callback could be used to initialize and notify things that needs to be in executed after UI rendering


muhammadusamaijaz:
> 

ok, I'll add an issue for it

PR #803, opened 2023-02-07
Moved super.dispose down

### LINKED ISSUE #802: Dispose order is wrong
**Describe the bug**
The order of the dispose method is wrong. 
Now, it first disposes the object and afterwards tries to remove the listeners, which doesn't make sense since it is already disposed. A package I use also throws an error if you execute code after disposing, which does make sense. 

[Flutter specification on state dispose](https://api.flutter.dev/flutter/widgets/State/dispose.html) is relevant here: Implementations of this method should **end** with a call to the inherited method, as in super.dispose().
I would strongly recommend adding this to the documentation and handling it the same. 

In short this:
```dart
  @override
  void dispose() {
    super.dispose();
    _removeListener?.call();
    if (_didInitValue) {
      delegate.dispose?.call(element!, _value as T);
    }
  }
```
should become this:
```dart
  @override
  void dispose() {
    _removeListener?.call();
    if (_didInitValue) {
      delegate.dispose?.call(element!, _value as T);
    }
    super.dispose();
  }
```

ikbendewilliam:
Fixes #802

### DIFF
--- a/test/null_safe/change_notifier_provider_test.dart
+++ b/test/null_safe/change_notifier_provider_test.dart
@@ -356,4 +356,61 @@ void main() {
 
     expect(myNotifier.notifyListeners, throwsAssertionError);
   });
+
+  testWidgets('call dispose after removing listeners', (tester) async {
+    final notifier = _ChangeNotifierDisposeTest();
+    final widget = ChangeNotifierProvider(
+      create: (_) => notifier,
+      builder: (context, __) => Text(
+        Provider.of<_ChangeNotifierDisposeTest>(context).text,
+        textDirection: TextDirection.ltr,
+      ),
+    );
+
+    await tester.pumpWidget(widget);
+    await tester.pumpAndSettle();
+
+    // If the listener is removed after the super.dispose, this will throw
+    await tester.pumpWidget(Container());
+    await tester.pumpAndSettle();
+  });
+}
+
+class _ChangeNotifierDisposeTest with _ChangeNotifierDispose {
+  final String text = 'test';
+}
+
+mixin _ChangeNotifierDispose implements ChangeNotifier {
+  var _disposed = false;
+  var _listeners = 0;
+
+  @override
+  void dispose() {
+    _disposed = true;
+    if (_listeners != 0) {
+      throw Exception('listeners not removed before dispose');
+    }
+  }
+
+  @override
+  void removeListener(VoidCallback listener) {
+    _listeners--;
+    if (_disposed) {
+      throw Exception('disposed when removing listener');
+    }
+  }
+
+  @override
+  void addListener(VoidCallback listener) {
+    _listeners++;
+    if (_disposed) {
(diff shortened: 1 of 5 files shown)


rrousselGit:
Hello!

Could you write tests for these changes?

ikbendewilliam:
I'm having trouble writing a good test, I can't seem to recreate the same issue as we have in our plugin. I did add a test that validates the delegate isn't disposed before the listeners have been removed though 🤔

test/null_safe/change_notifier_provider_test.dart
```diff
@@ -356,4 +356,61 @@ void main() {
...
+      builder: (context, __) => Text(
+        Provider.of<_ChangeNotifierDisposeTest>(context).text,
+        textDirection: TextDirection.ltr,
+      ),
+    );
+
+    await tester.pumpWidget(widget);
+    await tester.pumpAndSettle();
+
+    // If the listener is removed after the super.dispose, this will throw
```
rrousselGit, line 373: That's not the case. ChangeNotfier no-longer throws if removeListener is invoked post dispose.

The assert has been removed.

rrousselGit: Sorry by bad, I had yet to read the custom exception bit

rrousselGit:
@ikbendewilliam Did these changes fix the issue with your plugin?

rrousselGit (approved):
LGTM, but make sure it works on your app first. It'd be a shame if it didn't

ikbendewilliam:
@rrousselGit Turns out I'm sick, so when I'm feeling better I'll have another go at writing better tests and I'll confirm that everything is fixed for our plugin. Thank you for your patience

rrousselGit:
No problem! Thanks for your work :)

ikbendewilliam:
I have good news and bad news 😅 
The good news is that I'm able to reproduce it. The bad news is that I used two providers and the error I was receiving was not because the order of the dispose is wrong, but because `dispose` got called a second time. 
This branch obviously doesn't fix that since the issue is not in the package. 

My opinion, however, is still that the order should be first remove listeners and then super.dispose. So for me this can be merged, WDYT?

rrousselGit:
Honestly I'm a bit skeptical.
This could be breaking some folks, since that behavior has been around for ages now.

And the current behavior isn't problematic. You're the first to point it out, I never saw/heard of the issue before.
Removing the listeners _after_ dispose is generally harmless for most provided objects.

rrousselGit:
Hello again!

I don't think I will go through with these changes after-all. In the end:
- provider is in Long Term Maintenance mode. I avoid to avoid breaking changes
- this could be silently breaking
- the value is not that large
- people haven't requested for this before.

So I'll keep things as they are for now.
But this could be revisited if a large number of people ask for this


I appreciate your effort though!

snaeji:
We are in the same situation. It would be very beneficial for us if this would be merged so we don't have to fork 👍

PR #805, opened 2023-02-17
Explicitly provide type parameters.

eyebrowsoffire:
This is an info diagnostic that will soon become a warning.

See https://github.com/flutter/flutter/issues/120992


rrousselGit:
Thanks!
I don't mind these changes, but it's a bit unclear to me why they are needed.

PR #811, opened 2023-03-29
Added helpful notes for using Provider and Navigator

Mohamed-Shalabi:
### Added helpful notes for using Provider and Navigator
- It is preferred to keep the providers scoped as possible, not above the whole application.
- The ***Navigator.push*** methods create a new branch of the tree for the new screen under the MaterialApp, so the providers in first screen cannot be accessed in the second screen.
- I added two methods to deal with this situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,11 +226,27 @@ or the documentation of [Consumer](https://pub.dev/documentation/provider/latest
 and [Selector](https://pub.dev/documentation/provider/latest/provider/Selector-class.html)
 for more information.
 
-**NOTE**:
+**NOTES**:
 
-When using ***Navigator.push*** methods, a new branch of the widget tree is created for the new screen under the MaterialApp. So, you won't be able to access the Provider in the first screen.
+When using ***Navigator.push*** methods, a new branch of the widget tree is created for the new screen under the MaterialApp. So, you won't be able to access the Provider that is in the first screen from the second screen.
 
-It is recommended to keep the data scoped as possible, so you can send the instance to the new screen using `Provider.value`. It is recommended that if the second screen needs the provider to save data in it then you can await for ***Navigator.push*** and return this data through Navigator arguments.
+It is recommended to keep the data scoped as possible, so you can send the instance to the new screen using `Provider.value`.
+
+```dart
+return MaterialPageRoute(
+  builder: (BuildContext _) => ChangeNotifierProvider.value(
+    value: context.read<MyProvider>(),
+    child: const SecondScreen(),
+  ),
+);
+```
+
+It is recommended that if the second screen needs the provider to save data in it then you can await for ***Navigator.push*** and return this data through Navigator arguments.
+
+```dart
+final data = await Navigator.of(context).pushNamed(SecondScreen.routeName);
+context.read<MyProvider>().setData(data);
+```
 
 ### Optionally depending on a provider
 
--- a/README.md
+++ b/README.md
@@ -234,7 +234,7 @@ It is recommended to keep the data scoped as possible, so you can send the insta
 
 ```dart
 return MaterialPageRoute(
-  builder: (BuildContext _) => ChangeNotifierProvider.value(
+  builder: (_) => ChangeNotifierProvider.value(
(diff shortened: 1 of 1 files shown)


rrousselGit:
Hello!

> It is preferred to keep the providers scoped as possible, not above the whole application.

That isn't my recommendation, no.
I suggest putting everything above MyApp. It both makes testing easier and removes the concerns of navigation/context.

Mohamed-Shalabi:
Hi!

Thanks for your quick response.

I thought that this consumes more memory and that is bad from an
architecture aspect (I can access anything from everywhere).

But I didn't know your recommendation. Sorry for this misunderstanding.

rrousselGit:
Not scoping providers makes maintaining your app generally much simpler.  

Them all being accessible by widget isn't an issue. You shouldn't have logic in your UI anyway.
What matters is provider-to-provider interactions, which respect a unidirectional dataflow (a provider can only read the other providers above it).
So in that sense, the state isn't really accessible everywhere no; only by the UI.

Mohamed-Shalabi:
Oh that makes sense, thanks a lot.

Do you recommend any resources to learn how to make a good architecture for
the project?

‫في الخميس، 30 مارس 2023 في 1:43 ص تمت كتابة ما يلي بواسطة ‪Remi
Rousselet‬‏ ***@***.***‬‏>:‬

yimao009:
> ### Added helpful notes for using Provider and Navigator
> * It is preferred to keep the providers scoped as possible, not above the whole application.
> * The _**Navigator.push**_ methods create a new branch of the tree for the new screen under the MaterialApp, so the providers in first screen cannot be accessed in the second screen.
> * I added two methods to deal with this situation.

Hello,  
```dart
Navigator.of(context).push(
                      MaterialPageRoute(
                        builder: (_) => ChangeNotifierProvider.value(
                          value: context.read<Counter>(),
                          child: const SecondPage(),
                        ),
                      ),
                    );
```
I didn’t understand the comment you submitted. Can you provide a simple demo?
This transfers the value of the first page to the second page，

Mohamed-Shalabi:
Hi @yimao009 !
If you provided a change notifier (or anything else) over a screen, and navigated from this screen to another one -> the new screen is not put in the tree as a child to the first screen, it is put as a neighbor to the old screen under the `Navigator` widget (you can use Flutter DevTools to check widget tree after navigation).
So, you cannot access you change notifier using the context of the new screen.
this scenario also occurs when using something like `showModalBottomSheet`.
I hope you understood me!

yimao009:
```dart
return MaterialPageRoute(
  builder: (_) => ChangeNotifierProvider.value(
    value: context.read<MyProvider>(),
    child: const SecondScreen(),
  ),
	@@ -244,7 +244,7 @@ return MaterialPageRoute(
It is recommended that if the second screen needs the provider to save data in it then you can await for ***Navigator.push*** and return this data through Navigator arguments.

```dart
final data = await Navigator.of(context).pushNamed<MyDataClass>(SecondScreen.routeName);
context.read<MyProvider>().setData(data);
```

----
Can you provide a simple demo? I didn’t understand this comment.

yimao009:
return MaterialPageRoute(
  builder: (_) => ChangeNotifierProvider.value(
    value: context.read<MyProvider>(),
    child: const SecondScreen(),
  ),

---
Is this used to jump to the second page? follow code:

```dart
       Navigator.of(context).push(
                MaterialPageRoute(
                   builder: (_) => ChangeNotifierProvider.value(
                                               value: context.read<MyProvider>(),
                                               child: const SecondScreen(),
            )
      )

```
is that so？

yimao009:
> Hi @yimao009 ! If you provided a change notifier (or anything else) over a screen, and navigated from this screen to another one -> the new screen is not put in the tree as a child to the first screen, it is put as a neighbor to the old screen under the `Navigator` widget (you can use Flutter DevTools to check widget tree after navigation). So, you cannot access you change notifier using the context of the new screen. this scenario also occurs when using something like `showModalBottomSheet`. I hope you understood me!

叼毛，人呢

PR #827, opened 2023-07-12
Fix typos

HACCP92:
It means to relate to the facts or judgments of the latter section with vague questions. (어떤 값이던 간에 값을 노출시킵니다.(X) )

It means that there is no difference in which one is selected. (어떤 값이든 간에 값을 노출시킵니다.(O) )

(어떤 값이던 (X) => 어떤 값이든(O) )


rrousselGit:
Thanks!

PR #832, opened 2023-08-18
Create DevTools extension for `package:provider`

kenzieschmoll:
This sets up the skeleton for the DevTools extension though and moves the provider screen code from flutter/devtools to the provider repo.

**New**: Provider extension supplied by `package:provider`:

Screenshot 2023-08-18 at 2 28 23 PM

**Old**: Provider screen implemented as part of DevTools:

Screenshot 2023-08-18 at 2 28 37 PM

Draft for https://github.com/rrousselGit/provider/issues/833


rrousselGit:
Thanks for this! Is there anything specific I should look into? Maybe I'm expected to debug the error?

kenzieschmoll:
I just updated https://github.com/rrousselGit/provider/issues/833#issuecomment-1696495932 with an update on this work. With the latest commit, you should be able to pull down this PR and build in the simulated DevTools environment. Yes, if you could help with debugging the error and verifying this extension is WAI, that would be great. Other TODOs include writing a true integration test and ensuring the tests that we originally had in DevTools work here. 

Eager to hear your feedback on the development workflow and any feature requests.

rrousselGit:
Thanks for this!

Would it make sense & be reasonable to publish dev release of the extension with the current state of things? (after I've had the time to play around with it a bit :) )

kenzieschmoll:
> Would it make sense & be reasonable to publish dev release of the extension with the current state of things? (after I've had the time to play around with it a bit :)

Right now the extension support in DevTools is behind an experiments flag (hoping to turn that on by end of september), so even if you publish a dev version of provider with the extension, you wouldn't be able to load it in production DevTools (i.e. you'd still have to build DevTools locally with experiments enabled to see your extension in a real DevTools instance). 

We are also changing the `config.json` spec to be a `config.yaml` file (https://github.com/flutter/devtools/pull/6280), but I expect that change to land this week. All this to say, there is nothing that stops you from landing and publishing a dev version, but there may be some more breaking changes coming in the next few weeks. So my recommendation would be to hold off on publishing for a week or two.

rrousselGit:
Great, sounds good to me. Thanks for sharing :)

kenzieschmoll:
Hi @rrousselGit this is ready to go. With the latest commit, AIs 1) and 4) above should be resolved. I fixed 4) and I just haven't been able to reproduce 1). AIs 2) and 3) still stand.

The latest versions of `devtools_extensions` and `devtools_app_shared` have been published. Expect some polish to come over the next couple weeks but these are safe to start using. The latest commit I pushed up here bumps the pubspec to use the latest published versions. DevTools extensions support will be landed in the SDK shortly (I expect it to get to Flutter master within a couple days).

Until then, you can develop your extension locally with the Simulated DevTools environment, or build DevTools locally to test your extension against a real DevTools instance. Instructions for both are here: https://pub.dev/packages/devtools_extensions#dart--flutter-devtools-extensions.

Please let me know if you have any questions or how I can help. Thanks!

extension/devtools/build/flutter.js
```diff
@@ -0,0 +1,376 @@
+// Copyright 2014 The Flutter Authors. All rights reserved.
```
rrousselGit, line 1: There are lots of things in that `build` directory. Do we need them all? Can't we gitignore them and only keep what we need? :)

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -110,10 +111,15 @@ class ProviderScreenBody extends ConsumerWidget {
...
-  // bannerMessages.addMessage(
-  //   ProviderUnknownErrorBanner(screenId: ProviderScreen.id).build(),
-  // );
+  extensionManager.showBannerMessage(
+    key: 'provider_unknown_error',
+    type: 'error',
+    message: '''
+DevTools failed to connect with package:provider.
+
+This could be caused by an older version of package:provider; please make sure that you are using version >=5.0.0.''',
```
rrousselGit, line 120: Isn't the devtool version tied with the provider package version?

In which case, shouldn't they already be using the "up-to-date" version?

extension/devtools/build/flutter.js
rrousselGit: Oh I see, it's a build output of the extension.

Hum... I don't quite like seeing this here, and don't think other packages would either. It sounds like something that begs to be outdated, and pollutes the dev environment a bit.
How was this generated? `flutter build <path_to_extension>` followed by moving the generated build dir there? 

I'd probably ignore it and use a custom tool for publishing which would pre-compile the extension along with publishing

kenzieschmoll: Do the instructions I added at https://github.com/flutter/devtools/pull/6402 address your concerns?

packages/provider_devtools_extension/lib/src/provider_screen.dart
kenzieschmoll: This was just copied over from what the existing provider screen in DevTools had. Now that the provider screen (as a DevTools extension) is tied to the package:provider version is is shipped with, this is probably obsolete.

extension/devtools/build/flutter.js
rrousselGit: It explains briefly the problem yes.
Do you think it would be feasible to add a "pub publish --dry-run" validation step for this?

As in fail the validation if a devtool extension is detected but not built

kenzieschmoll: @jonasfj would this be possible? Perhaps this would be a sort of "pub extension" where you can add publish validation checks for a package. It seems like this sort of thing would require a pubspec change like we discussed early on. Otherwise I'm not sure where pub would know where to get the additional pub publish validations. 
```
  extensions:
    - devtools
```

Another idea is a more general field for adding custom publish validations
```
publish_checks:
  - devtools_extensions: ^1.0.0
    - verify_extension_build   # maybe this is an executable provided by package `devtools_extensions`?
```

kenzieschmoll:
@rrousselGit do you want to take this PR over? or do you want me to land it as, and then you can address [2) and 3)](https://github.com/rrousselGit/provider/pull/832/#issuecomment-1696500134)  in a follow up?

extension/devtools/build/flutter.js
jonasfj: Today, `pub publish --dry-run` validations are implemented in `pub`.

I would not oppose to hardcode in a check for `extensions/devtools/..`.

But it would be a one-off hack.

It would of course be awesome, if `pub` would find additional validations from `extension/pub/config.yaml` in dependencies of the package being published.

But there'd be security implications. Maybe we could prompt users for consent...

@sigurdm thoughts?

rrousselGit: I'm fine with hardcoding it too. I think that makes sense.

Extensibility for that is much less important IMO, and probably in the scope of the "plugins for Dart/Flutter CLI" instead of the devtool.

rrousselGit:
I was thinking about merging it as is, and dealing with tests/folders separately on my own.

Is it fine to make a dev release with this now? I assume we could make a dev release even if it isn't on Flutter's master channel. And it'd be accessible as soon as this lands on Flutter's side

extension/devtools/build/flutter.js
kenzieschmoll: A one-off hardcoded check for DevTools sgtm. We can always remove later and replace with a more extensible validation solution if we create one.

kenzieschmoll:
> I was thinking about merging it as is, and dealing with tests/folders separately on my own.
>
> Is it fine to make a dev release with this now? I assume we could make a dev release even if it isn't on Flutter's master channel. And it'd be accessible as soon as this lands on Flutter's side

Yes it is okay to make a dev release now. This plan SGTM. The extension support should be on Flutter master, and a more polished version of the extension support will be on master by the end of this week (I fixed a few bugs over the last few days I will release into the Dart SDK shortly).

extension/devtools/build/flutter.js
jonasfj: Discussing this offline, we're sort of wondering if maybe the number of users who will be publishing devtools extensions is so limited that maybe we shouldn't invest into having special checks for pre-publishing.

kenzieschmoll: If we expect the extensions ecosystem to grow, which is the goal, I still think it could be good to have a check. It doesn't have to be anything too complicated, but something simple that checks:
```
if the `extension/devtools/` directory exists
   verify the directory contains a file `config.yaml` and a non-empty `build/` directory
```
if would still be the responsibility of the extension author to ensure they are updating the contents of `build/` as necessary.

rrousselGit:
Alright, thanks for this!

I've made a dev release (6.1.0-dev.0) with those changes.

rrousselGit:
Ah, I gitignored the `build` folder. But it looks like this ignores it from `pub publish` too.
I've added a `pubignore` to circumvent this. It's likely worth mentioning in the docs next to the bit about not commiting `build`

rrousselGit:
Looks like all is well now. I've managed to use the extension on the example app using the published Provider version.

That's awesome!

extension/devtools/build/flutter.js
kenzieschmoll: Filed https://github.com/dart-lang/pub/issues/4026 to track this feature request.

PR #836, opened 2023-09-08
Update zh-CN/README.md

Apach3Q:
In the English documentation, the term 'Providers' is used to represent all Provider-related functionalities. However, in the Chinese documentation, the term used is 'Provider.' This might potentially create ambiguity for those who solely rely on the Chinese documentation and could mislead beginners into thinking that the `Provider` class can perform watch.

This pr highlights the distinction between the collective term "Providers" and the specific usage of the `Provider` class in both English and Chinese documentation.


rrousselGit:
Sounds good to me, thanks :)

Issue #830, opened 2023-08-14
Use `Selector` with Records instead of Tuple


jagomf:
In `Selector` docs, [current recommendation is to use `tuple` package](https://github.com/rrousselGit/provider/blob/9f66261ac6f0155d8d4a941147bbded36dc89e30/lib/src/selector.dart#L126).

Probably the best recommendation now would be to use native **Dart 3 Records**. Docs and example should be updated accordingly.

rrousselGit:
Fancy making a PR? :)

tienthai0205:
Hi @rrousselGit I'd like to take this issue, can you assign it to me? Thanks

rrousselGit:
Sure, feel free to make a PR

Issue #712, opened 2021-12-22
Improve Selector documentation


subzero911:
https://pub.dev/documentation/provider/latest/provider/Selector-class.html

You give a complex example with tuple at once. It looks difficult for beginners. Consider adding an example for 1 value before, like this:
```
Here's an example:

Selector<Foo, Bar>(
  selector: (_, foo) => foo.bar,  // will rebuild only when `bar` changes
  builder: (_, data, __) {
    return Text('${data.item}');
  }
)

To select multiple values without having to write a class... // Tuple example
```

PR #855, opened 2024-01-12
docs: change selector example from tuple to records

GameOnAnil:
1. This PR closes [830](https://github.com/rrousselGit/provider/issues/830): Change Selector example doc from tuple to dart Records as current recommendation is to use records. 
2. This PR also closes [712](https://github.com/rrousselGit/provider/issues/712) by adding more example for Selector

### DIFF
--- a/packages/provider/lib/src/selector.dart
+++ b/packages/provider/lib/src/selector.dart
@@ -119,6 +119,22 @@ class _Selector0State<T> extends SingleChildState<Selector0<T>> {
 /// As such, it `selector` should return either a collection ([List]/[Map]/[Set]/[Iterable])
 /// or a class that override `==`.
 ///
+/// Here's an example:
+///
+/// Example 1:
+///
+///```dart
+/// Selector<Foo, Bar>(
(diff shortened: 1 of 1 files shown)


packages/provider/lib/src/selector.dart
```diff
@@ -119,20 +119,37 @@ class _Selector0State<T> extends SingleChildState<Selector0<T>> {
...
 ///   }
 /// )
+///```
+///In this example `builder` will be called only when `foo.bar` changes.
+///
+/// Example 2:
+///
+///To select multiple values without having to write a class that implements `==`,
+///the easiest solution is to use "Records," available from Dart version 3.0.
+///For more information on Records, refer to the [records](https://dart.dev/language/records).
```
rrousselGit, line 140:
```suggestion
/// To select multiple values without having to write a class that implements `==`,
/// the easiest solution is to use "Records," available from Dart version 3.0.
// /For more information on Records, refer to the [records](https://dart.dev/language/records).
```

rrousselGit:
LGTM, thanks!

Issue #854, opened 2024-01-01
Listen to Changes from a Notifier


PeterAkande:
Just as Its easy to add a listener callback to the BlocConsumer in flutter bloc, It would be great is there is such callback In the Consumers and Selectors provided by Provider.

For example, a Consumer with a Listener can be defined as:
```dart
Consumer<SomeState>(
    listener: (state){},
    builder: (context, state, child){
        return SomeWidget();
     },
);
```

rrousselGit:
This had been requested before.
I haven't found a graceful way to do it quite yet.

Bloc does it because it has `Bloc.listen`. But there's no equivalent for providers.
The closest is `State.didChangeDependencies`, but it can be invoked for reasons other than a specific dependency change.

PeterAkande:
Did this as a work around that.

What do you think about it?

```dart

class ConsumerWithListener<T> extends StatefulWidget {
  final Widget? child;
  final void Function(T value) listener;
  final Widget Function(
    BuildContext context,
    T value,
    Widget? child,
  ) builder;

  const ConsumerWithListener({
    super.key,
    this.child,
    required this.listener,
    required this.builder,
  });

  @override
  State<ConsumerWithListener<T>> createState() =>
      _ConsumerWithListenerState<T>();
}

class _ConsumerWithListenerState<T> extends State<ConsumerWithListener<T>> {
  Timer? debounceTimer;

  @override
  Widget build(BuildContext context) {
    return Consumer<T>(
      builder: (context, T value, child) {
        if (debounceTimer?.isActive ?? false) debounceTimer?.cancel();

        /// The build Function can be called multiple Times for each notification
        /// But we don't want our listener to be called so many times
        /// So lets Debounce some unneeded calls.
        /// This is especially needed in cases where A Navigator.push is done in the
        /// Listen function.
        ///
        /// An error that is due to assert(!_debugLock) is thrown.
        debounceTimer = Timer(const Duration(milliseconds: 100), () {
          widget.listener.call(value);
        });

        return widget.builder.call(context, value, child);
      },
      child: widget.child,
    );
  }
}

```

codexsourav:
OK i Add This

ricardodalarme:
Hey @rrousselGit, I've been thinking about a ProviderListener and came up with a potential solution. To handle mutable notifiers, it uses a `value` selector to listen for specific data changes. I'd love to hear your thoughts!

```dart
class ProviderListener<T, V> extends StatefulWidget {
[...]

PR #857, opened 2024-01-28
add #854 Listener

codexsourav:
Listen to Changes from a Notifier #854 Add listener on Consumer


rrousselGit:
Hello!

While I appreciate the through, that approach doesn't work. The listener would be notified even if the value didn't change when the widget rebuilds.

PR #871, opened 2024-03-17
Introducing Provider.shared for Efficient SharedInstance Management

dadagov125:
### Problem

In the current implementation of the application with navigation between screens, there is a need to use the same instance (e.g., state management or service) on different screens, such as `ScreenA` and `ScreenB`. The approach that involves creating an instance at a level above `MaterialApp` results in the instance remaining in memory throughout the application's lifecycle, which is not optimal given its limited use.

Passing data between screens via navigation arguments is limited and not always suitable for working with complex data structures or in the context of declarative navigation.

### Solution

A proposed approach involves using a shared repository (`SharedInstances`) for temporarily storing and subsequently reusing instances. This repository will automatically delete unused instances, freeing up resources.

**Mechanism:**

1. On `ScreenA`, an instance is created and registered in `SharedInstances`.
2. When navigating to `ScreenB` and needing to use the same instance, the provider checks for the presence of the required instance in `SharedInstances`. If found, the instance is reused.
3. After `ScreenA` and `ScreenB` are removed from the widget tree, the corresponding instance is deleted from `SharedInstances`.

### Advantages

- Efficient memory usage by destroying instances when they are no longer needed.
- Avoidance of duplicating instances for different screens.
- Simplification of data management between screens through the reuse of instances.

### DIFF
--- a/packages/provider/lib/src/inherited_provider.dart
+++ b/packages/provider/lib/src/inherited_provider.dart
@@ -95,6 +95,22 @@ class InheritedProvider<T> extends SingleChildStatelessWidget {
         ),
         super(key: key, child: child);
 
+  InheritedProvider.sharedInstance({
+    Key? key,
+    required Create<T> createInstance,
+    Dispose<T>? dispose,
+    String? instanceKey,
+    bool? lazy,
+    this.builder,
+    Widget? child,
+  })  : _lazy = lazy,
+        _delegate = _SharedInstanceInheritedProvider(
+          create: createInstance,
+          dispose: dispose,
+          instanceKey: instanceKey,
+        ),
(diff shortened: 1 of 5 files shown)


rrousselGit:
Hello!

I appreciate your eagerness to contribute, but this is skipping steps.
Before making a PR, it is critical to first open an issue to discuss the proposed changes.

As it is, it is too early for me to review a PR, because I'm not sure I agree with the proposed API quite yet.

Do you mind opening an issue first, so we can move discussion there? I'll close this in the meantime, and we could reopen it later once this is sorted.

dadagov125:
Ok! I'll bring this up for discussion!

PR #874, opened 2024-04-19
Fix typos

choi88andys:
This version may also acceptable.
```dart
 runApp(
   Builder(builder: (context) {
     final value = context.watch<Movie?>();

     if (value == null) Text('no Movie found');
     return Text(value.title);
   }),
 );
```

</br>

But I think this version is more self-explanatory.
```dart
 runApp(
   Builder(builder: (context) {
     final movie = context.watch<Movie?>();

     if (movie == null) Text('no Movie found');
     return Text(movie.title);
   }),
 );
```


rrousselGit:
LGTM thanks!

PR #885, opened 2024-09-03
Added translated README in Italian

ilCatta:
Added translated README files in Italian to enhance accessibility for Italian-speaking users.


rrousselGit:
Thanks!

PR #869, opened 2024-03-08
Wrap dev tools extension screen with SelectionArea

guenth39:
By wrapping the dev tools screen with `SelectionArea` it's now possible to select text and that way copy the current values or names of providers.

@rrousselGit There is no guidance on what I have to do if I want to contribute. Please let me know if there is anything more to do to ship this or just do it by yourself, that's fine for me as well.

PR #891, opened 2024-09-24
Update devtools dependencies

### LINKED ISSUE #882: DevTools extension are not buildable with latest Flutter/Dart
**Describe the bug**
Hi!

I was interested in proceeding work on [#868](https://github.com/rrousselGit/provider/issues/868) to improve DevTools extension.

Here is my dev environment:

```
OS: macOS 14.3.1
Flutter: 3.22.2
Dart: 3.4.3
DevTools: 2.34.3
```

Dependencies of extension (which are rapidly developing as I see) needed to be updated because:
1. [vm_service](https://github.com/rrousselGit/provider/blob/master/packages/provider_devtools_extension/pubspec.yaml#L20) version range is too restrictive and dependency resolution failed: 
```
Because every version of integration_test from sdk depends on vm_service 14.2.1 and provider_devtools_extension depends on vm_service >=11.9.0 <14.0.0, integration_test from sdk is forbidden.
So, because provider_devtools_extension depends on integration_test from sdk, version solving failed.
```
2. `devtools_app_shared` is outdated a bit and breaks tests. ([Link](https://github.com/rrousselGit/provider/pull/869#issuecomment-1985592804))

...And couple other issues I experienced during updating:

1. Seems like [this deprecation](https://pub.dev/documentation/devtools_extensions/latest/devtools_extensions/DevToolsExtension/requiresRunningApplication.html) broke somehow backward compatibility and now it's plugin's responsibility to check connection, at least in "development" (with `simulated_chrome_environment` option)

I will provide my fork with all these updates a little bit later in this issue since it's kinda horrible at the moment and can not be merged in any way. But it could be handy with reproducing next issue, which is most problematic I think. 

Because of obligatory update of `vm_service` to `14.2.1` and up, core logic of [instance_providers.dart](https://github.com/rrousselGit/provider/blob/master/packages/provider_devtools_extension/lib/src/instance_viewer/instance_providers.dart#L397-L425) breaks with a lot of null-related issues (because there is a lot of force-unwraps).
If comment all `_parseFields` and just use empty lists, there are no exceptions. But no useful info in devtools too, of course, since no fields obtained from VM.

**To Reproduce**

1. Clone my fork on branch `null-exceptions`, [Link](https://github.com/Sameri11/provider/tree/null-exceptions)
2. Run example app (`provider/packages/provider/example`). You will get log like this:
```
...
A Dart VM Service on Chrome is available at: http://127.0.0.1:53606/jatJWHd2X30=
The Flutter DevTools debugger and profiler on Chrome is available at:
http://127.0.0.1:9101?uri=http://127.0.0.1:53606/jatJWHd2X30=
...
```

Obtain VM address from this log (in this case – ` http://127.0.0.1:53606/jatJWHd2X30=`)
3. Run extension example with `--dart-define=use_simulated_environment=true` as described [here](https://github.com/rrousselGit/provider/blob/master/packages/provider_devtools_extension/README.md)
4. Connect to Dart VM from step 2 in upper-right corner (`Dart VM service connection`)
5. Check error in `Logs` tab

**Expected behavior**

I was expecting [instance_providers.dart](https://github.com/Sameri11/provider/blob/null-exceptions/packages/provider_devtools_extension/lib/src/instance_viewer/instance_providers.dart#L397-L425) to work with updated `vm_service` and updated [VM service protocol](https://github.com/dart-lang/sdk/blob/main/runtime/vm/service/service.md)

Sameri11:
This PR is an attempt to adapt provider's devtools extension to modern version of dependencies, especially dependencies related to DevTools itself.

Main changes:
1. Extension is adapted to situation when `VmService` is not provided from the very start. Back then, this was impossible situation since extension (any extension, not just subj) was not standalone and was aware of connection. Now, it's extension's responsibility to monitor connection state. Because of that, it was very hard (or maybe even impossible) to work with `simulated_environment` as described [here](https://github.com/rrousselGit/provider/blob/master/packages/provider_devtools_extension/README.md)

2. Tests were broken as described in  https://github.com/rrousselGit/provider/pull/869#issuecomment-1985582204. My investigation led me to fact that `devtools_extensions` and other dependencies now depend on `dart:js_interop` and since this lib is imported, tests no longer can run on `vm` (as per https://github.com/flutter/flutter/issues/148670#issuecomment-2121963496). Hence, my attempt to adapt to this in this PR. Tests are runnable now with `flutter test --platform chrome` or through `vscode` config, which is also added in PR.

Overall, main purpose of changes is to allow and unblock further work on devtools extension.

Presumably fixes #882 and https://github.com/rrousselGit/provider/pull/869#issuecomment-1985582204

## Release Notes

- **New Features**
	- Introduced a new provider to monitor the connection state to the `VmService`, enhancing user feedback during connection attempts.
	- Enhanced functionality by replacing a placeholder annotation package with the official `freezed_annotation` package, improving data handling for instance details.
	- Introduced a loading state and error state in the UI for better responsiveness based on connection status.

- **Bug Fixes**
	- Enhanced error handling in the ProviderScreen, improving user feedback during connection issues.
	- Corrected string formatting in the `index.html` and `manifest.json` files.

- **Dependencies**
	- Updated several package dependencies to their latest versions for improved functionality and stability.

- **Tests**
	- Transitioned tests to a browser context, enhancing state management and UI validation for the ProviderScreen.
	- Added new tests to verify UI responses to various service connection scenarios.


Sameri11:
@rrousselGit Hello! Hope you're doing well.

I understand that this PR is quite extensive, and it might be challenging to review. Or, maybe, you already have your own plans for maintaining the extension, I completely respect that and will decline PR if this is the case.

However, I would love to get your feedback on this work if you have the time. Please let me know if you have any questions or if there's anything specific you’d like me to clarify. No pressure at all — your insights would be greatly appreciated!

packages/provider/example/.metadata
```diff
@@ -4,35 +4,35 @@
...
     - platform: web
-      create_revision: 8bc223e3e5cc919bada837ecc0afdc546969a79e
-      base_revision: 8bc223e3e5cc919bada837ecc0afdc546969a79e
+      create_revision: 761747bfc538b5af34aa0d3fac380f1bc331ec49
+      base_revision: 761747bfc538b5af34aa0d3fac380f1bc331ec49
     - platform: windows
-      create_revision: 8bc223e3e5cc919bada837ecc0afdc546969a79e
-      base_revision: 8bc223e3e5cc919bada837ecc0afdc546969a79e
+      create_revision: 761747bfc538b5af34aa0d3fac380f1bc331ec49
+      base_revision: 761747bfc538b5af34aa0d3fac380f1bc331ec49
```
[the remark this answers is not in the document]
Sameri11: Guess this file is out of scope completely, will remove this change from PR later.

packages/provider_devtools_extension/test/provider_screen_test.dart
```diff
@@ -2,28 +2,35 @@
...
+import 'package:devtools_app_shared/ui.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:flutter_test/flutter_test.dart';
-
+import 'package:vm_service/vm_service.dart';
 import 'test_utils.dart';
 
+class MockVmService extends Mock implements VmService {}
+
```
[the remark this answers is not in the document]
Sameri11: Was not sure about using mocks here at all, will address this a little bit later.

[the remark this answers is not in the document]
rrousselGit: Nah ignore this. I don't want to use code-generated mocks.

packages/provider_devtools_extension/lib/src/instance_viewer/fake_freezed_annotation.dart
Sameri11, line 1: Not sure about these changes though. Maybe, since it's standalone extension now, there is no much sense in avoiding dependencies like this comment says?
https://github.com/rrousselGit/provider/blob/64745784fcef9e3ebc6e6dea0d5f82d1ac20c797/packages/provider_devtools_extension/lib/src/instance_viewer/instance_details.dart#L13-L17

I mean, is it still necessary to avoid real freezed_annotations and builders etc here?

Sameri11: Missed this first time, caught with failed tests :)

rrousselGit: For sure! We should be using it now that it's a standalone :)
I haven't really done anything to the extension when it was extracted away.

Sameri11: Done!

packages/provider_devtools_extension/lib/src/instance_viewer/instance_providers.dart
```diff
@@ -406,7 +406,7 @@ Future<List<ObjectField>> _parseFields(
     final owner =
         await eval.safeGetClass(fieldDeclaration.owner! as ClassRef, isAlive);
 
-    final ownerUri = fieldDeclaration.location!.script!.uri!;
+    final ownerUri = owner.location?.script?.uri ?? '';
```
rrousselGit, line 409: Why this change?

packages/provider_devtools_extension/lib/src/instance_viewer/instance_viewer.dart
```diff
@@ -118,15 +120,18 @@ class _InstanceViewerState extends ConsumerState<InstanceViewer> {
...
   ) {
     if (error is SentinelException) {
       final valueAsString = error.sentinel.valueAsString;
       if (valueAsString != null) return [Text(valueAsString)];
     }
 
-    return const [Text('<unknown error>')];
+    extensionManager.postMessageToDevTools(DevToolsExtensionEvent(
+        DevToolsExtensionEventType.unknown,
+        data: {'stacktrace': stack.toString(), 'error': error.toString()}));
```
rrousselGit, line 133:
```suggestion
        data: {'stacktrace': stack.toString(), 'error': error.toString()}),);
```

packages/provider_devtools_extension/lib/src/instance_viewer/instance_viewer.dart
```diff
@@ -118,15 +120,18 @@ class _InstanceViewerState extends ConsumerState<InstanceViewer> {
...
+    return [Text('<unknown error>\n$stack')];
```
rrousselGit, line 134:
```suggestion

    return [Text('<unknown error>\n$stack')];
```

rrousselGit: I don't like the idea of showing the Stack like that. That's a multiline message on a single-line UI

And if we're going to show the stack, we should show the error.

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -12,19 +13,22 @@ import 'instance_viewer/instance_providers.dart';
 import 'instance_viewer/instance_viewer.dart';
 import 'provider_list.dart';
 import 'provider_nodes.dart';
+import 'instance_viewer/eval.dart';
 
-final _hasErrorProvider = Provider.autoDispose<bool>((ref) {
-  if (ref.watch(sortedProviderNodesProvider) is AsyncError) return true;
+final _hasErrorProvider = Provider.autoDispose<AsyncError?>((ref) {
+  if (ref.watch(sortedProviderNodesProvider) is AsyncError) {
+    return ref.read(sortedProviderNodesProvider) as AsyncError;
```
rrousselGit, line 20: You shouldn't use ref.read here.

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -12,19 +13,22 @@ import 'instance_viewer/instance_providers.dart';
 import 'instance_viewer/instance_viewer.dart';
 import 'provider_list.dart';
 import 'provider_nodes.dart';
+import 'instance_viewer/eval.dart';
 
-final _hasErrorProvider = Provider.autoDispose<bool>((ref) {
-  if (ref.watch(sortedProviderNodesProvider) is AsyncError) return true;
+final _hasErrorProvider = Provider.autoDispose<AsyncError?>((ref) {
```
rrousselGit, line 18: The name of the provider feels odd now.

And we probably shouldn't be returning AsyncError itself, but just the error/stack

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -12,19 +13,22 @@ import 'instance_viewer/instance_providers.dart';
...
 
-  if (selectedProviderId == null) return false;
+  if (selectedProviderId == null) return null;
 
   final instance = ref.watch(
     instanceProvider(InstancePath.fromProviderId(selectedProviderId)),
   );
 
-  return instance is AsyncError;
+  return instance is AsyncError ? instance as AsyncError : null;
```
rrousselGit, line 31: This shouldn't need a cast.

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -42,71 +46,97 @@ class ProviderScreenBody extends ConsumerWidget {
...
         : '[No provider selected]';
 
-    ref.listen<bool>(_hasErrorProvider, (_, hasError) {
-      if (hasError) showProviderErrorBanner();
+    ref.listen<AsyncError?>(_hasErrorProvider, (_, hasError) {
+      if (hasError != null) {
+        extensionManager.postMessageToDevTools(
+            DevToolsExtensionEvent(DevToolsExtensionEventType.unknown, data: {
+          'stacktrace': hasError.stackTrace.toString(),
+          'error': hasError.toString()
```
rrousselGit, line 62:
```suggestion
          'error': hasError.toString(),
```

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -42,71 +46,97 @@ class ProviderScreenBody extends ConsumerWidget {
...
     // A provider will automatically be selected as soon as one is detected
     final selectedProviderId = ref.watch(selectedProviderIdProvider);
     final detailsTitleText = selectedProviderId != null
         ? ref.watch(_selectedProviderNode)?.type ?? ''
         : '[No provider selected]';
 
-    ref.listen<bool>(_hasErrorProvider, (_, hasError) {
-      if (hasError) showProviderErrorBanner();
+    ref.listen<AsyncError?>(_hasErrorProvider, (_, hasError) {
+      if (hasError != null) {
```
rrousselGit, line 58: As I said above, the provider name makes no sense to me anymore. So the value needs to be renamed too.

We don't have an "hasError" value but the raw error itself.

packages/provider_devtools_extension/lib/src/instance_viewer/instance_viewer.dart
```diff
@@ -118,15 +120,18 @@ class _InstanceViewerState extends ConsumerState<InstanceViewer> {
...
+    StackTrace? stack,
     InstancePath __,
   ) {
     if (error is SentinelException) {
       final valueAsString = error.sentinel.valueAsString;
       if (valueAsString != null) return [Text(valueAsString)];
     }
 
-    return const [Text('<unknown error>')];
+    extensionManager.postMessageToDevTools(DevToolsExtensionEvent(
```
rrousselGit, line 131: We shouldn't post messages inside widgets.

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -42,71 +46,97 @@ class ProviderScreenBody extends ConsumerWidget {
...
+          'error': hasError.toString()
+        }));
+        showProviderErrorBanner();
+      }
+    });
+    serviceManager.connectedState.addListener(() {
+      if (serviceManager.connectedState.value.connected) {
+        setServiceConnectionForProviderScreen(serviceManager.service!);
+      }
     });
```
rrousselGit, line 71: Listeners cannot be added inside widgets like this. You'd be causing a memory leak and accumulate more and more listeners over time.

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -42,71 +46,97 @@ class ProviderScreenBody extends ConsumerWidget {
...
-                          context: context,
-                          builder: (_) => _StateInspectorSettingsDialog(),
-                        ),
-                      );
-                    },
+    // This change needed for extension to work in simulated environment
+    // since new option `requiresConnection` to config.yml
+    // not works in this config and it is needed to check connection
+    // inside plugin itself.
+    return ref.watch(serviceProvider).when(
```
rrousselGit, line 77: Don't use`when`. Use `switch` with pattern matching

packages/provider_devtools_extension/lib/src/provider_screen.dart
```diff
@@ -42,71 +46,97 @@ class ProviderScreenBody extends ConsumerWidget {
...
-          ),
-        ),
-      ],
-    );
+              ],
+            ),
+        error: (_, stacktrace) => Center(
+              child: Text(
+                  'Error when connecting to VmService: \n${stacktrace.toString()}'),
+            ),
```
rrousselGit, line 136: I don't think there's value in showing that in the UI. It probably should be a snackbar or something.

packages/provider_devtools_extension/pubspec.yaml
```diff
@@ -11,13 +11,13 @@ dependencies:
...
-  # TODO: https://github.com/flutter/devtools/issues/4569 - unpin this version
-  flutter_riverpod: 2.0.0-dev.9
+  collection: ^1.18.0
+  devtools_extensions: ^0.2.2
+  devtools_app_shared: ^0.2.2
+  flutter_riverpod: ^2.5.1
   logging: ^1.1.1
-  vm_service: ">=11.9.0 <14.0.0"
+  vm_service: ^14.2.5
+  freezed_annotation: ^2.4.4
```
rrousselGit, line 20: Nice!

packages/provider_devtools_extension/lib/src/instance_viewer/instance_providers.dart
Sameri11: Yeah, great question, should've take notes for myself :( But, story is like this: `fieldDeclaration` (`field.decl`) is deprecated field, and I thought it's great idea to start getting rid of it's usages. Apart of this, while I was testing, it was always null. Plus, it's kinda strange to gather location of field itself here, and not owner – I understand that they should not differ in this case, but from semantics point view it seems odd for me.

But! I will reconsider this part anyway, will come back with proper explanation at least.

rrousselGit: First, IMO switching from `!.` to `?.   ?? ''` isn't right. `''` isn't a valid URI and it's just going to silently fail later.
We should fail early.


We do care about where the field is located, not where the owner is located. An owner and a field may be in a different location (as a field may be inherited by a superclass or a mixin).

The logic here is trying to write code to read a field. As such, it is mandatory to run the read operation in the context of where the field is defined. This is for one core reason: The field may be private.

---

Ultimately, the variable name is likely wrong here. It should be `fieldUri` not `ownerUri`.

Sameri11:
Ok, I decided to cut the scope a little here and simplify couple things:

1. About [this comment](https://github.com/rrousselGit/provider/pull/891#discussion_r1782568629) – reverted completely because apparently old code works flawlessly. Guess I was misleaded by another errors here.
2. About all changes in error reporting – reverted completely. Couldn't justify them even for myself, so decided not to change anything at all. So yeah, maybe it worth separate issue if it's desired to focus on this.
3. Simplified work with `serviceManager` and connection detection. 

Also, tried to address all comments (regarding wrong usage of providers and leaks etc) – hence completely new commit. I admit that previous work was rather horrible and hope that this attempt is better.

PR #912, opened 2025-06-13
Add dispose to FutureProvider

RossComputerGuy:
This is useful for disposing data which was created in a future provider.


packages/provider/lib/src/async_provider.dart
```diff
@@ -181,10 +181,15 @@ class FutureProvider<T> extends DeferredInheritedProvider<Future<T>?, T> {
...
   /// `create` must not be `null`.
+  ///
+  /// The optional [dispose] callback is invoked with the last exposed value
+  /// when the provider is removed from the widget tree, allowing cleanup of
+  /// resources (closing streams, cancelling timers, etc.).
   FutureProvider({
     Key? key,
     required Create<Future<T>?> create,
     required T initialData,
+    Dispose<T>? dispose,
```
rrousselGit, line 192: This is impractical because more often than not, it's not `T` that needs disposing when using futures/streams. It's something else, like a Completer/StreamController/...

RossComputerGuy: Oh, the problem I was running into was that if it were a `Future` then I didn't have access to the data since it was already completed. How would I get around this to work for `Completer` or a `StreamController`?

